PR #25, opened 2015-07-31
New schema, test data, docs, moved "legacy" tests into proper categories.

hjellinek:
Updated to use latest schemas, which resulted in many small (and a few larger) changes.
Updated tests and docs.  The ctk-java (tests) module now produces clean JavaDoc.
Wrote some new tests, e.g. for Datasets.
Added a set of standard data to test against, `compliance-dataset1`.
Fleshed out the example test classes in the `cts-demo-java` module.
Removed `LegacyComplianceIT.java` after moving its tests to appropriate homes under `api.{reads,references,variants}`.


macieksmuga:
We agreed that merging into the ga4gh:compliance_redux branch doesn't require the usual complement of +1's. They will be asked for when we request to merge this branch into master.

PR #38, opened 2015-08-15
Skip test

wstidolph:
I adopted a policy of "don't do target server integration tests on every build" so that I could work on the CTK itself more quickly. To do that, I added a property `cts.skipITs` in parent/pom.xml and use it to configure the cts-java/pom.xml to skip tests.You set a command line property or environment variable to actually run the tests (so, e.g., `mvn install -Dcts.skipITs=false`) I also update the RunningTests_maven doc to add this property, and catch it up a bit closer to current state.


diekhans:
Given the CTK is a server development tool, testing it against a
server should really default to false.

Instead of using a property, could the install and test targets
be independent?

wstidolph:
Personally I default to "skipITs=true" because I make a lot more CTK
changes than server changes :) but it's an easy default to set
individually. Should I set to false (run server test by default) and
resubmit PR?

Splitting the targets is the "right" thing to do but not really easy; it
probably needs a GA4GH-specific maven plugin, and perhaps a custom maven
lifecycle definition. (Of course, there would be ancillary benefits...)

diekhans:
It should default to not running the tests.

Our own Maven plugin, that is depressing.   Could we
have the test divided up into those that don't talk to
the server and a special target for those that do?

XML declarative programming is the devil incarnate.

wstidolph:
The problem with having a separate target is less the XML than it is the fundamental maven "lifecycle" model ... which is a design topic well beyond the scope of this PR :)

As it stands, this PR defaults to not running the tests when running maven builds, so I'll leave the PR as-is. Not that this only affects *maven*use ... if you run the command-line tool (which is what I assumed a *server* developer would mostly do), it runs the server tests.

wstidolph:
Whoa, I'm ulling this PR, I have no idea what I did to cause it to affect 200+ files! My bad, I thought I affects *3* files, so I did something very very wrong ...

PR #39, opened 2015-08-16
Don't run server tests by default (cts.skipITs support)

wstidolph:
(Re-submitting PR, based on compliance_redux branch this time)

I want to "skip target server integration tests" on most builds - it slows down work on the CTK itself, particularly when I'm not running a GA4GH server at the default server location so there's test failures and timeouts.

To do that without adding a command line property on every run, or having to set an environment property, I added a property `cts.skipITs` in `parent/pom.xml` and use that to configure the `cts-java/pom.xml` failsafe plugin for whether or not to skip the cts-java tests that talk to the target server.

Then, when you *do* want to run the tests, you set a command line property  (so, e.g., `mvn install -Dcts.skipITs=false`)

If you're server or server-test developer and you want to run the tests every time you invoke mvn install, just invert the property in parent/pom.xml.

I also updated the RunningTests_maven.md doc to show use of this property, and to catch the doc up a bit closer to current state.

### DIFF
--- a/cts-java/pom.xml
+++ b/cts-java/pom.xml
@@ -148,6 +148,7 @@
                             <goal>verify</goal>
                         </goals>
                         <configuration>
+                            <skipITs>${cts.skipITs}</skipITs>
                             <argLine>-Dctk.tgt.urlRoot=${ctk.tgt.urlRoot}</argLine>
                             <properties>
                                 <property>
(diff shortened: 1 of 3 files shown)


hjellinek:
I added the feature and docs (edited to suit) in a PR I just merged.  That was easier than merging this one.

PR #57, opened 2015-08-24
New coherent test data set.

macieksmuga:
This PR replaces the prior ad-hoc datasets with a single coherent dataset based on selected small genes against the GRCh37 reference. It consists of:

* One VCF file per gene (currently only BRCA1 is represented) containing calls from the gene's region for a few selected samples.
* SAM files (one per sample) containing aligned reads for the selected samples in the gene's region.
* One FASTA file representing the reference GRCh37 assembly in the region of the gene. 

All files have their coordinates offset to the start of the region, so that the three file types (FASTA, SAM, VCF) are on the same local coordinate system.


jeromekelleher:
This looks great, thanks @macieksmuga!

+1.

hjellinek:
An important step.

+1.

PR #59, opened 2015-08-25
More concrete recipes for binary conversion of data.

macieksmuga:
Per @hjellinek's request, concrete commands used to convert the human-readable data into binary formats as used by the reference server.


diekhans:
Please do these are build rules rather than documentation.

If maven can't do it use, just use make.

macieksmuga:
@diekhans I believe PR #60 satisfies your request - are there any remaining blocks to merging this PR in?

Issue #37, opened 2015-08-14
complaince report should include query and response messages.


diekhans:
From: Jerome Kelleher
I agree with @calbach that it would be very helpful for the HTML test reports to contain the
actual messages transferred. This was a genuinely useful part of the old compliance tests.

david4096:
Took a good look at this today. It's feasible and certainly useful. For some of the tests hundreds of megabytes are transferred, so these logs can be quite long! There is a little bit of coaxing the JUnit reporter into displaying the results in a useful way.

david4096:
I have a first draft of this feature here: https://github.com/david4096/compliance/tree/trafficlogs

It currently uses the approach of naming traffic log files after their respective tests. By customizing the XSLT for generating reports I'm able to link to the traffic logs for each test separately. Sometimes the traffic logs can be quite long!

There are a few edge cases that make this approach problematic: when tests have the same name across namespaces or when there are special characters in the file name.

PR #62, opened 2015-08-29
Act info svcs

wstidolph:
(Apologies this PR got so big, I was working on it while "ownership" of compliance moved and now it's catch-up time. I've tracked it against the compliance_redux branch up until about 8/25/2015 so I'm hoping by the time this PR is evaluated it won't be *too* far behind and difficult to merge! )

## GOALS:
1. Add services and refactoring to enable "test coverage" test (zzCheckCoverage) and writing tests that look across the test run, or that need information about what objects/types are defined in the IDL.
2. Add multi-root test classpath and application-dev profile (better flexibility in using IDE/maven).
3. Miscellaneous changes in logging control, and scattered cleanup of comments/javadoc.

## CHANGES BY CATEGORY:
* add TrafficLog domain object (structured object to replace formatted log string)
* added TrafficLogService (and repository) to store and serve TrafficLogs as record of "what happened in the test run." The storage can be a static Map in the Service, or if you run under Spring (as in the CLI and Server modes) then the Service uses an embedded H2 DB.
* add a 'run key' for isolating traffic logs from different runs in the repository
* add future-use field 'testmethodkey' to TrafficLog for when there are multiple server interactions in a single @Test method (not yet setting/using the field, just got it into the data definition)
* add DomainInformationService (collect IDL-based data types during build)
* modify zzCheckCoverage to add tests comparing IDL-defined data (data types, URL endpoints) against used-in-testing data
* refactoring as ctk-domain grew: split into org.ga4gh.ctk.{domain, services, utility} subpackages and moved "Props" into "utility" to prevent ctk-domain (domain support) having dependency on testrunner - that's backwards, left over cruft
* added specific Loggers for the new {domain, services, utility} in log4j2.xml in both ctk-cli and ctk-server, and generally added comments/ordering to clarify log4j2.xml files. 
* support multi-root test classes so you can have the test jar but override some of it with test classes, or just have the test classes pick up on your IDE's target classes directory. All the test classes are packaged up into a single tests jar at start of ant run, and that jar is kept in the test results/<tgt url>/<testsequence number>/lib in case someone wants to keep the exact test classes that caused some observed test result. Impact on application.properties and Props (add new control property), and in antRunTest.xml.
* restructured antRunTest.xml for clarity and to better pass groups of properties
* add application-dev.properties so you can launch with 'dev' spring profile active under an IDE and not have to create fake 'lib' dirs in ctk-cli and ctk-server
* add service use/description notes to WritingATest.md
* javadoc, comment fixups as I encountered them


calbach:
Is this PR still relevant? Can it be dropped?

wstidolph:
Personally, I think the ideas in the PR are still relevant - I don't think they've been independently realized. But the code in 'compliance' has moved on since I filed the PR so now it would need conflict resolution to be merged. 

After this is merged (or independently re-coded) it would be easy to use the improved TrafficLog etc to capture the sent/received JSON for a failing test and have it available for working on issue #37 (which needs the data available during the report-writing phase, after all the tests have run).

hjellinek:
Closing it for now.  We should consider implementing these features/ideas from scratch, given enough demand for them.

PR #80, opened 2015-09-15
Fix the test data reference set checksum

calbach:
This is computed as `MD5("90977a37195d3fd247e4916b5b4cbae8")` per the [documentation](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/references.avdl#L89).


hjellinek:
Assigning to compliance dataset wrangler @macieksmuga .

calbach:
@macieksmuga is this good to merge? I'm mainly concerned that this implies a bug in the reference server (I'm assuming the current version passes against the reference server), so it would be good to propagate that information back.

macieksmuga:
@calbach yes, this is good. Sorry I didn't get around to it before.

PR #82, opened 2015-09-15
Change expected refset taxon ID from string -> int

calbach:
This field has type `{null, int}` in the [schema](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/references.avdl#L69).


jeromekelleher:
+1.

calbach:
I'm just going to merge this one since it seems uncontroversial and I'm pretty sure the current tests which use the field are guaranteed to fail (given they check equality of Integer and String). Also Jerome's +1 implies there is no conflict with the reference server.

Issue #53, opened 2015-08-19
Should not expect multi-readGroup search to be UNIMPLEMENTED


calbach:
I suspect this is a case of overfitting to the reference server.

ReadMethodsEndpointAliveIT
```
Expecting responseStatus of: <org.ga4gh.ctk.transport.WireTracker@4269e2df> to be: <NOT_IMPLEMENTED> but was: <null>
```

PR #102, opened 2015-09-29
Adding back reads tests.

macieksmuga:


cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsGetByIdIT.java
```diff
@@ -0,0 +1,53 @@
...
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void testGetByIdMatchesSearch() throws AvroRemoteException {
+        final SearchReadGroupSetsRequest request =
+                SearchReadGroupSetsRequest.newBuilder()
+                                          .setDatasetId(TestData.getDatasetId())
+                                          .build();
+
+        // search for all ReadGroupSets
```
calbach: This is slightly misleading because you haven't paginated and therefore there is no guarantee you've gotten all read group sets.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
```diff
@@ -0,0 +1,136 @@
...
+                                              .build();
+            final SearchReadGroupSetsResponse pageResp = client.reads.searchReadGroupSets(pageReq);
+            final List<ReadGroupSet> pageOfReadGroupSets = pageResp.getReadGroupSets();
+            pageToken = pageResp.getNextPageToken();
+
+            assertThat(pageOfReadGroupSets).hasSize(1);
+            setOfReadGroupSetsGathered1By1.add(pageOfReadGroupSets.get(0));
+        } while (pageToken != null);
+
+        assertThat(setOfReadGroupSetsGathered1By1).containsAll(setOfExpectedReadGroupSets);
```
calbach: This is a weaker assertion than what you want, I think. For instance, gathered1By1 could be a superset.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    @Test
+    public void checkPagingOneByOneThroughReadGroupSets() throws AvroRemoteException {
+
+        // retrieve them all - this may return fewer than "all."
+        final List<ReadGroupSet> listOfReadGroupSets = Utils.getAllReadGroupSets(client);
+        assertThat(listOfReadGroupSets).isNotEmpty();
+
+        // we will do a set comparison after retrieving them 1 at a time
+        final Set<ReadGroupSet> setOfExpectedReadGroupSets = new HashSet<>(listOfReadGroupSets);
```
calbach: Maybe no action needed: It would be nice if you could avoid using a set here altogether, but I guess the spec doesn't say anything about the return order of the results. A sane implementation should return results in the same order regardless of page size (I'm not sure that needs to be explicitly stated in the spec).

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
```diff
@@ -0,0 +1,136 @@
...
+                                              .setDatasetId(TestData.getDatasetId())
+                                              .setPageSize(1)
+                                              .setPageToken(pageToken)
+                                              .build();
+            final SearchReadGroupSetsResponse pageResp = client.reads.searchReadGroupSets(pageReq);
+            final List<ReadGroupSet> pageOfReadGroupSets = pageResp.getReadGroupSets();
+            pageToken = pageResp.getNextPageToken();
+
+            assertThat(pageOfReadGroupSets).hasSize(1);
+            setOfReadGroupSetsGathered1By1.add(pageOfReadGroupSets.get(0));
```
calbach: It would be good to validate that you aren't adding the same read group set twice to your set. Using a list instead of a set would sidestep this issue nicely.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsSearchIT.java
```diff
@@ -0,0 +1,214 @@
...
+
+    private static Client client = new Client(URLMAPPING.getInstance());
+
+    /**
+     * Return the number of Strings that match the possible substring.
+     * @param possibleSubstring the substring we're searching for
+     * @param matchWithin the {@link String[]} we're searching within
+     * @return the number of elements of <tt>matchWithin</tt> that contain <tt>possibleSubstring</tt>
+     */
+    private static int countMatches(String possibleSubstring, String[] matchWithin) {
```
calbach: Unused?

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,179 @@
...
+        // then do a read search
+        final SearchReadsRequest srr =
+                SearchReadsRequest.newBuilder()
+                                  .setReadGroupIds(aSingle(readGroupId))
+                                  .setStart(0L)
+                                  .setEnd(150L)
+                                  .setReferenceId(refId)
+                                  .build();
+        final SearchReadsResponse rtn = client.reads.searchReads(srr);
+        assertThat(rtn.getAlignments()).isNullOrEmpty();
```
calbach: Why was this the conclusion? We've sent a query on 0-150 for the 'first valid reference' (order is not defined, AFAIK), and we expect a null or empty result?

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Check that an unmatched {@link SearchReadsRequest} results in a thrown exception.
+     *
+     * @param readGroupId the ReadGroup ID, passed in using the @Parameters annotation
+     * @throws Exception the exception
+     */
+    @SuppressWarnings("ThrowableResultOfMethodCallIgnored")
+    @Test
+    @Parameters({
+            "myNonsenseId", "foo", ""
```
calbach: Not sure I agree with the third case. I think the server is within its rights to throw a 400 invalid argument if you don't supply a readGroupId at all. I feel this is nearly equivalent to passing a 'null'  or 'undefined' in JSON for the readGroupId parameter.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,179 @@
...
+    // The workaround is simple enough, and actually seems a bit more readable,
+    // so we'll pass in a (valid) key to a static Map (see readGroupIdMap just below),
+    // and in the test method we'll look up the actual parameter.
+    @Parameters({
+            "TWO_GOOD, NOT_IMPLEMENTED",
+            "ONE_GOOD_ONE_BAD, NOT_IMPLEMENTED",
+            "TWO_BAD, NOT_IMPLEMENTED",
+            "THREE_GOOD, NOT_IMPLEMENTED"
+    })
+    public void multipleReadGroupsNotSupported(String readGroupId, RespCode expStatus) throws Exception {
```
calbach: Please remove this test per #53

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,179 @@
...
+
+        final SearchReadsRequest srr =
+                SearchReadsRequest.newBuilder()
+                                  .setReadGroupIds(Collections.emptyList())
+                                  .setStart(0L)
+                                  .setEnd(150L)
+                                  .setReferenceId(refId)
+                                  .build();
+        final GAWrapperException t = catchGAWrapperException(() -> client.reads.searchReads(srr));
+        assertThat(t.getHttpStatusCode()).isEqualTo(HttpURLConnection.HTTP_NOT_IMPLEMENTED);
```
calbach: Why would we return not implemented here? It's just not part of the spec.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
diekhans: the documentation indicates that at least one readgroup is
required, so I would interpret this as a invalid argument if
omitted.  However, it should be explict in the documentation.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        final List<ReadGroupSet> readGroupSets = resp.getReadGroupSets();
+
+        final SearchReadsRequest srReq =
+                SearchReadsRequest.newBuilder()
+                                  .setReferenceId(Utils.getValidReferenceId(client))
+                                  .setReadGroupIds(aSingle(Utils.getReadGroupId(client)))
+                                  .setStart(emptyRangeStart)
+                                  .setEnd(emptyRangeEnd)
+                                  .build();
+        // XXX fails due to server issue #591
```
calbach: This issue is closed. Remove?

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        final long end = 160;
+
+        final SearchReadsRequest srReq =
+                SearchReadsRequest.newBuilder()
+                                  .setReferenceId(Utils.getValidReferenceId(client))
+                                  .setReadGroupIds(aSingle(Utils.getReadGroupId(client)))
+                                  .setStart(start)
+                                  .setEnd(end)
+                                  .build();
+        // XXX fails due to server issue #591
```
calbach: This issue has been closed.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        final String readGroupId = Utils.getReadGroupId(client);
+
+        final SearchReadsRequest request =
+                SearchReadsRequest.newBuilder()
+                                  .setReadGroupIds(aSingle(readGroupId))
+                                  .setStart(0L)
+                                  .setEnd(150L)
+                                  .setReferenceId(refId)
+                                  .build();
+        // XXX fails due to server issue #591
```
calbach: Issue is closed.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        final SearchReadsRequest request =
+                SearchReadsRequest.newBuilder()
+                                  .setReadGroupIds(Collections.emptyList())
+                                  .setStart(0L)
+                                  .setEnd(150L)
+                                  .setReferenceId(refId)
+                                  .build();
+
+        final GAWrapperException t = catchGAWrapperException(() -> client.reads.searchReads(request));
+        assertThat(t.getHttpStatusCode()).isEqualTo(HttpURLConnection.HTTP_NOT_IMPLEMENTED);
```
calbach: This is a duplicate of the other test (see other comment for why I don't agree with the behavior).

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        response.getAlignments().stream()
+                .forEach(readAlignment ->
+                                 assertThat(readAlignment.getAlignedSequence()).isNotNull()
+                                                                               .matches("[ACTGN]+"));
+    }
+
+    /**
+     * Verify that passing all known read group names in a {@link SearchReadsRequest}
+     * returns all matching read groups.
+     * XXX fails due to server issue #596
```
calbach: Does this belong here? I generally read 'XXX' as 'resolve this issue before submitting'.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+
+        response.getAlignments().stream()
+                .forEach(readAlignment ->
+                                 assertThat(readAlignment.getAlignedSequence()).isNotNull()
+                                                                               .matches("[ACTGN]+"));
+    }
+
+    /**
+     * <p>Verify aligned sequences contain only the permitted symbols.</p>
+     * <p>In any {@link ReadAlignment}, the <tt>alignedSequence</tt> field can only contain
```
calbach: There is no such requirement in the schemas, as this comment appears to imply. However, the test suite controls the input data, so you could assert this as a property of the input data.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -0,0 +1,263 @@
...
+        for (String readGroupName : TestData.EXPECTED_READGROUP_NAMES) {
+            final String readGroupID = Utils.getReadGroupIdForName(client, readGroupName);
+            final SearchReadsRequest request =
+                    SearchReadsRequest.newBuilder()
+                                      .setReadGroupIds(aSingle(readGroupID))
+                                      .setReferenceId(refId)
+                                      .setStart(0L)
+                                      .setEnd(150L)
+                                      .build();
+            // XXX fails due to server issue #591
```
calbach: rm

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
hjellinek: As I read [the sparse documentation](http://joel-costigliola.github.io/assertj/core/api/index.html?org/assertj/core/api/AbstractIterableAssert.html), AssertJ's `assertThat(A).containsAll(B)` tests that `A` is a superset of `B`.  Which is what's wanted, correct?

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
hjellinek: Leaving aside questions of sanity :-), you are correct, there are no guarantees regarding ordering.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
hjellinek: We could also test if the set size increases by 1 with each attempted addition to the set.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsSearchIT.java
hjellinek: Correct, vestigial.  I will remove it.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
hjellinek: Looks like #53 needs attention.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: Yes, I just went through and removed all of these notes-to-myself.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: Gone.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: Gone.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: Not anymore.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: You are correct.  I will update the comment in this (ancient) test.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: Yes.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: I respect that.  I had updated the schemas `documentation` branch to document this behavior, since it was previously utterly undefined and undocumented.  Please feel free to open an issue against the schemas to covert this; the compliance suite can be made to test for any agreed-upon behavior, of course.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
hjellinek: I documented it in the schemas documentation branch based on what I thought was consensus regarding the behavior - I guess it wasn't consensus after all!

Please feel free to open an issue against schemas.  We have to document this behavior.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
calbach: On my first read through I assumed that `Utils.getAllReadGroupSets()` gets all read group sets. Does it really only get the first page? I think the functionality should change or it should be renamed.

If it doesn't get all ReadGroupSets, then what we're testing here is:
"Getting the first page of read group sets with the default page size returns a subset of the results yielded by paginating through all read group sets one by one".

I feel like a simpler and more consistent test case would show that "using the default page size yields the same results as using a page size of 1".

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
calbach: Thanks, I commented on https://github.com/ga4gh/schemas/pull/422.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
calbach: Please remove one of the two duplicate tests.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
hjellinek: This method dates to the dawn of time.  As such, it's now history.  Or will shortly be.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
hjellinek: It (and the other `getAll*` methods) should probably be renamed to something like `getTheDefaultNumberOf*`.  They were named before I learned that the server is not required to return all results when you don't pass in a `pageSize`, but can return any number of results (> 0) up to the server's maximum (an internal constant).  This is defferent from having a default page size, I think.  And makes for weaker tests than I would have liked.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
hjellinek: I removed the one in `ReadMethodsEndpointAliveIT`.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
calbach: By default, I meant a server default (which is not necessarily constant). Upon further thought, I guess this test still does something useful (though maybe too dynamic for my tastes) in that it asserts that an unpaginated set of results is somewhat consistent with a paginated set.

I feel the utility should definitely be renamed.

hjellinek: Since these utilities are used by the various compliance test branches, I'd prefer to do any such renaming after we merge the tests into master.

calbach: Sounds good to me.

hjellinek: Thinking out loud: I can imagine two sets of utility methods: one batch that truly does return all instances of the given object type, and another that returns the "first batch" of results.  The current methods fit in the latter category.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
calbach: I say we remove the third case. If we make the spec stricter on this point, we can add corresponding compliance test coverage.

diekhans: Please file a ticket on making the spec clear and add the Documentation label.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadGroupSetsPagingIT.java
jeromekelleher: I think this is something we should file as an issue to follow up on, but to my eye there's a lot of duplicated code testing the same functionality in terms of paging. I've made the point in the references tests as well. A simpler and more readable approach would be to make the page size an option in the client, and to provide methods that take the search method parameters directly (getting rid of the process of building requests entirely from these test cases). 

We should also have tests that provide malformed search parameters, of course. But for testing output of searching by name and so on, a method that returns a list/iterator would be a lot more useful than working with requests and doing paging every time.

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadMethodsEndpointAliveIT.java
calbach: Discussed offline: This is probably fine for now.

calbach:
+1

hjellinek:
+1

PR #103, opened 2015-09-29
Adding back variants tests.

macieksmuga:


cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
```diff
@@ -0,0 +1,139 @@
...
+     * using an increment as large as the non-paged set of results.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkPagingByOneChunkThroughCallSets() throws AvroRemoteException {
+        final String variantSetId = Utils.getVariantSetId(client);
+        final List<CallSet> listOfCallSets = Utils.getAllCallSets(client, variantSetId);
+
+        // page through the variants in one gulp
```
sarahhunt: /variants/callSets

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
```diff
@@ -0,0 +1,139 @@
...
+        checkSinglePageOfCallSets(variantSetId,
+                                  listOfCallSets.size(),
+                                  listOfCallSets);
+    }
+
+    /**
+     * Check that we can page through the variants we receive from
```
sarahhunt: /variants/callSets/

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
```diff
@@ -0,0 +1,150 @@
...
+        final List<CallSet> listOfCallSets = Utils.getAllCallSets(client, variantSetId);
+        assertThat(listOfCallSets).isNotEmpty();
+
+        // we will do a set comparison after retrieving them 1 at a time
+        final Set<CallSet> setOfExpectedCallSets = new HashSet<>(listOfCallSets);
+        assertThat(listOfCallSets).hasSameSizeAs(setOfExpectedCallSets);
+
+        final Set<CallSet> setOfCallSetsGathered1By1 = new HashSet<>(setOfExpectedCallSets.size());
+
+        // page through the ReferenceSets using the same query parameters and collect them
```
sarahhunt: /ReferenceSets/CallSets/

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsGetByIdIT.java
```diff
@@ -0,0 +1,57 @@
...
+        final String variantSetId = Utils.getVariantSetId(client);
+        final List<Variant> variants = Utils.getAllVariantsInRange(client, variantSetId, start, end);
+
+        assertThat(variants).hasSize(expectedNumberOfVariants);
+
+        for (final Variant variantFromSearch : variants) {
+            final Variant variantFromGet = client.variants.getVariant(variantFromSearch.getId());
+            assertThat(variantFromGet).isNotNull();
+
+            // XXX fails due to server issue #619
```
sarahhunt: Is this comment needed?

sarahhunt:
I looked at the layout of the Variants tests to get an idea how I would organise the VariantAnnotation tests and what were the standard aspects being tested, but they are somewhat inconsistent.

Seeing VariantsSearchIT and VariantSetsSearchIT, I expect to see CallSetsSearchIT rather than CallsetsSearchResponseCheckIT.
There are GET tests for Variants and CallSets,  but not for VariantSets. The Variants GET test is seperate to the the the Variants POST tests but the CallSet GET test is with the POST tests. 

It is not clear that there is a standard format to follow though it would help users and developers understand the suite quicker if there was.

That said, the tests run against our updated endpoints and give results I expect, so I would not want to delay synchronisation, so I am happy to +1.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
```diff
@@ -0,0 +1,150 @@
...
+
+    private static Client client = new Client(urls);
+
+    /**
+     * Check that we can page 1 by 1 through the {@link CallSet}s we receive from
+     * {@link org.ga4gh.ctk.transport.protocols.Client.Variants#searchCallSets(SearchCallSetsRequest)}.
+     * <p>
+     * The call to retrieve all {@link CallSet}s may return fewer than all of them, subject to
+     * server-imposed limits.  The 1-by-1 paging must enumerate them all, however.  The set of "all"
+     * must be a subset of those gathered one-by-one.
```
calbach: Per my comment on the Reads tests, why a subset and not equal?

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsGetByIdIT.java
hjellinek: No. :-)

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
```diff
@@ -0,0 +1,150 @@
...
+     * Check that we can page through the CallSets we receive from
+     * {@link org.ga4gh.ctk.transport.protocols.Client.Variants#searchCallSets
+     * (SearchCallSetsRequest)}
+     * using an increment twice as large as the non-paged set of results.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link
+     * GAException})
+     */
+    @Test
+    public void checkPagingByOneTooLargeChunkThroughCallSets() throws AvroRemoteException {
```
calbach: This test implicitly asserts that default page returned by the server includes all results. As discussed previously (and as expressed in the comment @ line 51) this is not necessarily true.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
hjellinek: Fixed.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
hjellinek: Because `Utils.getAllCallSets` works by passing `pageSize` = `null` in the `SearchCallSetsRequest`, and that's not guaranteed to return all `CallSet`s.  (Had I known that when first writing these tests, they'd likely be somewhat different.)  The strongest guarantee we can make is that the result of `Utils.getAllCallSets` will contain at least one `CallSet`, and that its `CallSet`s must be contained in the set we collect 1-by-1, since it's guaranteed to collect all `CallSet`s.

I interpreted your earlier comments on the other paging tests to take exactly this tack; did I misunderstand?

calbach: Sorry, I guess we can keep discussion on the other
comment thread for now. This is another case of me misinterpreting the behavior of `Utils.getAll*()`.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
calbach: We have the same issue in the other tests as well.

macieksmuga:
@jeromekelleher this should also be good to merge, assuming the +1's in the variants tests were intended across all the tests PRs.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Verify basic operations on variants, variantsets, and callsets.
+ */
+@Category(VariantsTests.class)
+public class VariantsMethodsEndpointAliveIT implements CtkLogs {
+
+    private static Client client = new Client(URLMAPPING.getInstance());
+
+    /**
+     * Check that searching {@link Variant}s with a nonexistent reference name returns
```
calbach: I don't necessarily agree that this is the right behavior. I could imagine this returning a 404 if you ask for a reference name which doesn't exist.

A reference which belongs to the associated reference set, but for which there is no associated data, would be less controversial.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,71 @@
...
+    public void testSearchCallSetsForNonexistentVariantSetFails() throws Exception {
+        final SearchCallSetsRequest scsr =
+                SearchCallSetsRequest.newBuilder()
+                                     .setVariantSetId(Utils.randomId())
+                                     .build();
+
+        final GAWrapperException gae =
+                Utils.catchGAWrapperException(() -> client.variants.searchCallSets(scsr));
+
+        assertThat(gae.getHttpStatusCode()).isEqualTo(HttpURLConnection.HTTP_NOT_FOUND);
```
calbach: Please handle the case where `gae` is null.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
```diff
@@ -0,0 +1,71 @@
...
+     */
+    @SuppressWarnings("ThrowableResultOfMethodCallIgnored")
+    @Test
+    public void testSearchCallSetsForNonexistentVariantSetFails() throws Exception {
+        final SearchCallSetsRequest scsr =
+                SearchCallSetsRequest.newBuilder()
+                                     .setVariantSetId(Utils.randomId())
+                                     .build();
+
+        final GAWrapperException gae =
```
calbach: Above you name this variable `t`

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+        final String variantSetId = Utils.getVariantSetId(client);
+        final List<Variant> listOfVariants = Utils.getAllVariantsInRange(client, variantSetId, start, end);
+
+        // we will remove Variants from this Set and assert at the end that we have zero
+        final Set<Variant> setOfVariants = new HashSet<>(listOfVariants);
+        assertThat(listOfVariants).hasSize(setOfVariants.size());
+
+        // page through the variants using the same query parameters
+        String pageToken = null;
+        for (Variant ignored : listOfVariants) {
```
calbach: This test is written as if start/end were an exact match query - it's actually an [overlap query](https://github.com/ga4gh/schemas/blob/master/src/main/resources/avro/variantmethods.avdl#L84) though.

Either this test doesn't work, or it happens to work because no two variants happen to overlap in the test data. If it's the latter, I don't think it makes sense to depend on that.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+    private static Client client = new Client(URLMAPPING.getInstance());
+
+    /**
+     * Check that we can page 1 by 1 through the variants we receive from
+     * {@link org.ga4gh.ctk.transport.protocols.Client.Variants#searchVariants(SearchVariantsRequest)}.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkPagingOneByOneThroughCalls() throws AvroRemoteException {
```
calbach: It's paging through Variants

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsSearchIT.java
```diff
@@ -0,0 +1,167 @@
...
+        assertThat(searchVariants).hasSize(expectedNumberOfVariants);
+    }
+
+    /**
+     * Check that we receive the expected number of {@link VariantSet}s.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkExpectedVariantSets() throws AvroRemoteException {
```
calbach: What is this test doing here?

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsSearchIT.java
```diff
@@ -0,0 +1,167 @@
...
+     * search contain well-formed {@link Call}s.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkCallsForWellFormedness() throws AvroRemoteException {
+        final long start = 50;
+        final long end = 100;
+
+        final SearchVariantSetsRequest req =
```
calbach: This should probably be extracted to a helper / Utils, it's repeated in most of these test cases.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallsetsSearchResponseCheckIT.java
```diff
@@ -0,0 +1,113 @@
...
+     */
+    @SuppressWarnings("ThrowableResultOfMethodCallIgnored")
+    @Test
+    public void getCallSetWithInvalidIDShouldFail() throws AvroRemoteException {
+        final String nonexistentCallSetId = Utils.randomId();
+
+        // fetch the CallSet with that ID
+        final GAWrapperException t =
+                Utils.catchGAWrapperException(() -> client.variants.getCallSet(nonexistentCallSetId));
+        assertThat(t.getHttpStatusCode()).isEqualTo(HttpURLConnection.HTTP_NOT_FOUND);
```
calbach: Please handle the case where `t` is null.

calbach:
Sorry, I was holding off on some indication that Sarah's comments had been addressed before taking another look. I added some minor comments - the only real significant issue I think is the variant pagination test.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsSearchIT.java
diekhans: Although please don't create anything called helpers/utils; move
common code to someplace functionally logical.  Dumping ground
modules don't make it easy to understand code (e.g. the
common.avdl, which almost none of it is common to the API).

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
diekhans: Please make sure that all the overlap corner cases are handled.
Things like:
  - overlap start only, end only
  - overlap middle
  - overlap exact match
  - overlap extends beyond
  - one base overlap
  - directly adjacent but non-overlapping

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsPagingIT.java
hjellinek: I've made changes to the reads-tests branch to deal with this, and will make matching changes here.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
hjellinek: The agreed (but not documented?) behavior, as I understand it, is that nonexistent IDs return a 404; nonexistent (non-matching) names filter by the name - and in this case, return an empty list.  I don't have a pointer to the discussion handy, unfortunately.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
hjellinek: It can't be null.  Check the implementation of `catchGAWrapperException`: it assumes (asserts) that the enclosed expression will throw an exception, and if it does not, it calls `fail()`, which causes control to transfer out of the entire test.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
hjellinek: This needs a rewrite to match the way we're handling paging in the other branches.  The tip-off is the `for (... ignored : ...)` idiom, which I've gotten rid of elsewhere.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallsetsSearchResponseCheckIT.java
hjellinek: As mentioned above, `t` cannot be `null`.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsSearchIT.java
hjellinek: I moved it.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
hjellinek: Renamed them.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsSearchIT.java
hjellinek: I created a pair of handy `map`-like methods in the same file, where they logically belong.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsMethodsEndpointAliveIT.java
calbach: I recall the thread you're referring to, and I believe this was in the context of a freeform metadata search for SearchReadGroupSets etc. This case is subtly different - `referenceName` is being used as an alias for a `Reference` within the associated `ReferenceSet`. This could debatably be switched with `referenceId`, or augmented with the option to pass a `referenceId` instead in which case you surely would expect a 404 for a non-existent `Reference`.

In the other case of freeform criteria search, `name` does not refer to any existing resource - it's just filtering on an attribute on the target resource.

I think we can call the specification ambiguous at best on this point, and I don't think the current behavior is a conservative assumption so I'd prefer this test were just removed for now. FYI: Google's API will pass this test case as you have it today, though I personally would prefer that it didn't.

cts-java/src/test/java/org/ga4gh/cts/api/variants/VariantsPagingIT.java
calbach: I think I may have misread the test on my first pass. I think this is fine (at least, it just has the same issues as the other pagination tests).

cts-java/src/test/java/org/ga4gh/cts/api/Utils.java
```diff
@@ -193,19 +192,29 @@ public static String getReadGroupIdForName(Client client, String name) throws Av
...
-                        .build();
-        final SearchReadGroupSetsResponse readGroupSetsResp =
-                client.reads.searchReadGroupSets(readGroupSetsReq);
-        assertThat(readGroupSetsResp).isNotNull();
-        final List<ReadGroupSet> readGroupSets = readGroupSetsResp.getReadGroupSets();
-        assertThat(readGroupSets).isNotEmpty().isNotNull();
-        return readGroupSets;
-    }
 
+        final List<ReadGroupSet> result = new LinkedList<>();
```
calbach: This can probably be refactored to be a call to `getAllReadGroupSetsForName(client, null)`.

calbach:
+1, just one more minor code cleanup comment.

cts-java/src/test/java/org/ga4gh/cts/api/Utils.java
calbach: Nevermind, read groups vs read group sets.

hjellinek:
Much obliged, @calbach!

PR #104, opened 2015-09-29
Adding back references tests.

macieksmuga:


cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+                                                                       .build();
+        final ListReferenceBasesResponse resp = client.references.getReferenceBases(refId, req);
+        return resp.getSequence();
+    }
+
+    /**
+     * Check that we can request a sequence of bases from
+     * {@link org.ga4gh.ctk.transport.protocols.Client.References#getReferenceBases(String,
+     * ListReferenceBasesRequest)}
+     * using a size twice as large as the full sequence, and receive the full sequence.
```
calbach: This doc does not appear to agree with the actual behavior.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+        final ListReferenceBasesRequest req = ListReferenceBasesRequest.newBuilder()
+                .setStart(sequenceStart).setEnd(sequenceEnd)
+                .build();
+        final ListReferenceBasesResponse resp = client.references.getReferenceBases(refId, req);
+        final String expectedSequence = resp.getSequence();
+        assertThat(expectedSequence).hasSize((int)sequenceEnd);
+
+        // walk through the bases by incrementing start and end
+        long offset = 0L;
+        for (char expectedChar : expectedSequence.toCharArray()) {
```
calbach: 150 serial requests seems like overkill. Perhaps a smaller range? Or larger intervals?

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+        final String fullSequence = getFullSequence();
+        checkChunkOfBases(TestData.REFERENCE_BRCA1_LENGTH, fullSequence);
+    }
+
+    /**
+     * Check that if we request a sequence of bases from
+     * {@link org.ga4gh.ctk.transport.protocols.Client.References#getReferenceBases(String,
+     * ListReferenceBasesRequest)}
+     * using a size larger than the full sequence, it fails with an exception with HTTP status
+     * "Requested Range Not Satisfiable."
```
calbach: This seems like reasonable behavior, but is it part of the spec?

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesSearchIT.java
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Retrieve all {@link Reference} objects for the default {@link ReferenceSet}. Check
+     * that each returned {@link Reference} has a valid-looking MD5.  (In the future, perhaps this
+     * test should compute the MD5 and compare it to what's returned.)
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkMD5ChecksumAppearanceOfAllReferences() throws AvroRemoteException {
```
calbach: This doesn't appear to be related to ReferenceBases, which is the name of the test.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
```diff
@@ -0,0 +1,132 @@
...
+        final List<ReferenceSet> listOfReferenceSets = Utils.getAllReferenceSets(client);
+        assertThat(listOfReferenceSets).isNotEmpty();
+
+        // we will remove ReferenceSets from this Set and assert at the end that we have zero
+        final Set<ReferenceSet> setOfReferenceSets = new HashSet<>(listOfReferenceSets);
+        assertThat(listOfReferenceSets).hasSize(setOfReferenceSets.size());
+
+        // page through the ReferenceSets using the same query parameters
+        String pageToken = null;
+        for (ReferenceSet ignored : listOfReferenceSets) {
```
calbach: This pagination pattern is repeated in a few places. Consider the test output when the endpoint returns fewer results than expected. This will result in having us loop back around to the start and emitting confusing assertion failures.

Perhaps accumulate all results into a set and perform a set difference at the end. Or else, assert that pageToken is non-null if there are still more expected reference sets.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
```diff
@@ -0,0 +1,132 @@
...
+     */
+    @Test
+    public void checkPagingOneByOneThroughReferenceSets() throws AvroRemoteException {
+
+        // retrieve them all
+        final List<ReferenceSet> listOfReferenceSets = Utils.getAllReferenceSets(client);
+        assertThat(listOfReferenceSets).isNotEmpty();
+
+        // we will remove ReferenceSets from this Set and assert at the end that we have zero
+        final Set<ReferenceSet> setOfReferenceSets = new HashSet<>(listOfReferenceSets);
```
calbach: I think naming this expectedReferenceSets would make things clearer.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Check that we can page through the {@link ReferenceSet}s we receive from
+     * {@link Client.References#searchReferenceSets(SearchReferenceSetsRequest)}
+     * using an increment as large as the non-paged set of results.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkPagingByOneChunkThroughReferenceSets() throws AvroRemoteException {
+
```
jeromekelleher: This is unlikely to fail in practise, but the server is not *required* to satisfy the pageSize requests. It's perfectly entitled to send back the results in pages of 1, if it so desires. This won't fail now because we only have a handful of objects, but if there was 1000 reference sets, then some servers would send back the results in chunks smaller than 1000. As such, I think this is a bit problematic and should possibly be removed.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * Check that we can page through the {@link ReferenceSet}s we receive from
+     * {@link Client.References#searchReferenceSets(SearchReferenceSetsRequest)}
+     * using an increment twice as large as the non-paged set of results.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkPagingByOneTooLargeChunkThroughReferenceSets() throws AvroRemoteException {
```
jeromekelleher: This is similar. All we can assert here is that we should get back all the results after paging. The sizes of the pages are up to the server, and anything > 0 is a valid value.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsSearchIT.java
```diff
@@ -0,0 +1,149 @@
...
+    @Test
+    public void checkSearchingReferenceSetsReturnsWellFormed() throws AvroRemoteException {
+        final SearchReferenceSetsRequest req =
+                SearchReferenceSetsRequest.newBuilder().build();
+        final SearchReferenceSetsResponse resp = client.references.searchReferenceSets(req);
+
+        final List<ReferenceSet> sets = resp.getReferenceSets();
+
+        sets.stream().forEach(rs -> assertThat(rs.getId()).isNotNull());
+        sets.stream().forEach(rs -> assertThat(rs.getMd5checksum()).isNotNull());
```
jeromekelleher: Should we be checking that this looks like an MD5 also?

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsSearchIT.java
```diff
@@ -0,0 +1,149 @@
...
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkReferenceSetFoundByAccession() throws AvroRemoteException {
+        final SearchReferenceSetsRequest req =
+                SearchReferenceSetsRequest.newBuilder()
+                        .setAccessions(TestData.REFERENCESET_ACCESSIONS)
+                        .build();
+        final SearchReferenceSetsResponse resp = client.references.searchReferenceSets(req);
+
```
jeromekelleher: These 3 methods are almost identical. The checking ``resp`` bit should be factored out.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferencesPagingIT.java
```diff
@@ -0,0 +1,149 @@
...
+        final List<Reference> listOfReferences = Utils.getAllReferences(client, refSetId);
+        assertThat(listOfReferences).isNotEmpty();
+
+        // we will remove References from this Set and assert at the end that we have zero
+        final Set<Reference> setOfReferences = new HashSet<>(listOfReferences);
+        assertThat(listOfReferences).hasSize(setOfReferences.size());
+
+        // page through the References using the same query parameters
+        String pageToken = null;
+        for (Reference ignored : listOfReferences) {
```
jeromekelleher: This seems quite low-level. Why not provide a method to do the paging automatically, which you can provide an optional page size (which I think @calbach suggests above too)? There's going to be an awful lot of code repetition if this is done for every object, on every search. I would say this test should be replaced with something like
```java
final Set<Reference> expectedReferences = Set(Utils.getAllReferences(client, refSetId));
final Set<Reference> foundReferences = Set(getReferences(client, refSetId, pageSize=1));
assertThat(expectedReferences == foundReferences);
```
(apologies for the mangled pseudocode, but you get the idea!)

The point is, we can't really every assert anything interesting about pageSizes other than they are > 0, so there's no point in checking these things.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferencesSearchIT.java
```diff
@@ -0,0 +1,149 @@
...
+                        .setReferenceSetId(referenceSetId)
+                        .build();
+        final SearchReferencesResponse resp = client.references.searchReferences(req);
+
+        final List<Reference> sets = resp.getReferences();
+
+        sets.stream().forEach(rs -> assertThat(rs.getId()).isNotNull());
+        sets.stream().forEach(rs -> assertThat(rs.getName()).isNotNull());
+        sets.stream().forEach(rs -> assertThat(rs.getLength()).isNotNull());
+        sets.stream().forEach(rs -> assertThat(rs.getMd5checksum()).isNotNull());
```
jeromekelleher: Looks like MD5?

jeromekelleher:
This looks great, thanks @hjellinek and @macieksmuga. The code is easy to read and the tests are very nice. I  agree with all the comments @calbach makes above.

I'm slightly worried about these paging tests though. I think we should aim to centralise the paging code and concentrate on testing the stream of objects we get back. This is the interesting bit.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
hjellinek: You're right, I fixed the comment.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
hjellinek: Not yet, but see ga4gh/schemas#430.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
macieksmuga: @jeromekelleher to be precise, my understanding is that a compliant server should never send back _more_ objects per request than specified by `pageSize` (it can certainly send less).

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
hjellinek: It's now, or will shortly be, `setOfExpectedReferenceSets`.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
hjellinek: I had indeed ignored that case; I plead ignorance.  Is this documented?  If not, please open an issue against the schemas.  You can even assign it to me! :-)

I have updated the `checkPagingOneByOneThroughReferences` and `checkPagingOneByOneThroughReferenceSets` methods similarly to what you describe.

calbach: My comment only pertains to the behavior of this specific test case, perhaps I could have written it more clearly. The reason it would loop back through is because you would set `pageToken` to null, effectively sending the same original request again.

In the scenario I described, the server is misbehaving in that it returns fewer results than expected - but the test should expect this as a possible failure scenario and should be clear in its failure message.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
jeromekelleher: Fair enough @macieksmuga --- we must have 0 < returnedPageSize <= requestedPageSize.

jeromekelleher:
Any updates on this @hjellinek?

macieksmuga:
With the pull of @hjellinek's latest code, are people's issues answered? Can we get the +1's to make this mergeable?

calbach:
I still see a bunch of unaddressed comments here (mostly from @jeromekelleher), I also just noticed one more issue.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
```diff
@@ -0,0 +1,152 @@
...
+     *
+     * @param chunkSize        the sequence length we'll request
+     * @param expectedSequence the full base sequence we expect
+     * @throws AvroRemoteException if there's a communication problem or server exception
+     */
+    private void checkChunkOfBases(long chunkSize,
+                                   String expectedSequence)
+            throws AvroRemoteException {
+
+        final String refId = Utils.getValidReferenceId(client);
```
calbach: This approach only works because there's exactly one `Reference` in the system. How I expected this to work ideally, is that a helper would perform a lookup by assembly ID to find your known reference set, then do a lookup by reference set ID + reference name to get a specific reference ID.

hjellinek: Good one.  There's a lot of room for new references-related tests, and the surest route to that would be to add more references to the compliance data set.  It's difficult to write tests with generality or confidence when you know there's only a single instance of the object under test.

Once there are multiple references, I expect things like `Utils.getValidReferenceId(client)` to go away in most tests in favor of the approach you outlined.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsSearchIT.java
hjellinek: I had a test like that previously, and class Utils already includes a method called `looksLikeValidMd5`.  I can add a call to that, but I think it would be better to add a check that the MD5 value is correct.  It's a much stronger test.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsPagingIT.java
hjellinek: I believe all the paging tests now check this.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceSetsSearchIT.java
hjellinek: Yup.  My philosophy, adapted from Martin Fowler: write unoptimized code, then refactor when you see a pattern or redundancy.  Or when @jeromekelleher does. :-)

Fixed, thanks.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferencesSearchIT.java
hjellinek: Can we not do _equals the correct MD5_, @macieksmuga?

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
hjellinek: I had missed this batch of @jeromekelleher's comments, but believe I have now addressed them.

calbach: The test has a hidden dependency on there being exactly one reference, which I consider to be an issue now - not only later, once more references are added. Ideally this would be fixed before the test is submitted, but if not there should at least be a TODO here (I wouldn't block my +1 on it being fixed here).

jeromekelleher:
OK, happy to +1 (once #116 is merged?). We could do with some more refactoring to take out the reliance of looping over pageTokens, but this is a good start.

hjellinek:
Thanks, @jeromekelleher.  I don't know that there's anything blocking #116, so I hope this can happen soon.

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesPagingIT.java
hjellinek: Thanks, @calbach. I just opened issues #117 and #118.  Issues are less likely to be overlooked than a TODO in the source code.

jeromekelleher:
I think you can just update this PR by pushing to it directly @hjellinek. Having PRs against PRs seems like overkill...

cts-java/src/test/java/org/ga4gh/cts/api/references/ReferenceBasesSearchIT.java
calbach: This is still unaddressed, but I won't block on it. I guess you could say the MD5 field of the references is technically related to the reference bases, though this seems like a bit of a stretch.

calbach:
+1, though I second the need for an eventual second pass on pagination here. There are still several places where the logic is not correct, though is likely to pass on most servers (e.g. all the tests which check pageSize * 2). It may make sense to pair a PR of pagination cleanup/fixes with some additional pagination documentation in the schemas branch.

jeromekelleher:
Also +1, and completely agree with @calbach's comments.

jeromekelleher:
I think we're ready to merge here. One question though: are we going to start squashing commits on PRs like this? Here we have some fairly boring updates and merge noise which would be much cleaner as a single commit. I'd prefer a clean history, but I guess it doesn't matter so much here in the compliance tests since we're not going to be making bugfix releases on older versions (probably??) so there won't be much cherry picking going on.

macieksmuga:
I'll be happy to squish the commits here and in the other parallel PR's

jeromekelleher:
Can we squash please @macieksmuga? Ready to merge otherwise I think.

macieksmuga:
Squashed commits & rebased, @jeromekelleher. Should be ready to merge.

jeromekelleher:
OK, we've got the requisite +1-age and a nice clean commit history, so I'm merging. Thanks @macieksmuga and @hjellinek!

Issue #88, opened 2015-09-17
'endpoints' tests do not appear to have any basis in the ga4gh/schemas spec


calbach:
These tests look to be written based on specific implementation details of the reference server, I would not expect them to pass on any other implementation. For example, a GET request to /callsets/search is interpreted by the Google API as a request for a callset with id 'search'.

To me, these looks more like server integration tests which I would expect to live in the reference server repository, if anywhere.

jeromekelleher:
I agree with @calbach on this example. What are the tests checking for?

hjellinek:
At @jeromekelleher's suggestion, I pared down the tests to a single one that checks the response to malformed POST payloads sent to search methods.

The one other useful test had been checking the response to `OPTIONS`, `GET`, and `POST` methods.  Arguably the schemas should specify the correct response to OPTIONS, at least, and test for it.

hjellinek:
Closed pending merge.

PR #105, opened 2015-09-29
Adding back endpoints tests.

macieksmuga:


macieksmuga:
I added these back as a separate PR to be consistent with the other test-adding PR's, but note that issue #88 questions the need for these tests as part of the compliance suite.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
```diff
@@ -0,0 +1,123 @@
...
+    private static String makeFullUrl(String partialUrl) {
+        return TransportUtils.makeUrl(urls.getUrlRoot(), partialUrl);
+    }
+
+    /**
+     * Test that the basic server verbs/methods work as expected.
+     *
+     * @param fullUrl the URL to test
+     */
+    private void testHttpMethods(String fullUrl) throws UnirestException {
```
jeromekelleher: This test is quite opaque to me --- what are we trying to test here?

This might be Java heresy, but it is helpful to break things up into multiple lines sometimes...

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
```diff
@@ -0,0 +1,123 @@
...
+        assertThat(Unirest.get(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_BAD_METHOD);
+        assertThat(Unirest.options(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_OK);
+        assertThat(Unirest.post(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_UNSUPPORTED_TYPE);
+    }
+
+    /**
+     * Test the status codes we're supposed to receive from the GET, POST, and OPTIONS methods on
+     * <tt>/callsets/search</tt>.
+     *
+     * <p>See server/tests/unit/test_views.py: testRouteCallsets</p>
```
jeromekelleher: I don't think we should be citing the server tests, if for no other reason that we refactor pretty aggressively and this file might not exist at all in a few months time.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
```diff
@@ -0,0 +1,123 @@
...
+
+    /**
+     * Test that the basic server verbs/methods work as expected.
+     *
+     * @param fullUrl the URL to test
+     */
+    private void testHttpMethods(String fullUrl) throws UnirestException {
+        assertThat(Unirest.get(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_BAD_METHOD);
+        assertThat(Unirest.options(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_OK);
+        assertThat(Unirest.post(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_UNSUPPORTED_TYPE);
```
jeromekelleher: Why is POST returning HTTP_UNSUPPORTED_TYPE here?

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Test the status codes we're supposed to receive from the GET, POST, and OPTIONS methods on
+     * <tt>/callsets/search</tt>.
+     *
+     * <p>See server/tests/unit/test_views.py: testRouteCallsets</p>
+     *
+     * @throws Exception if there's a connection problem
+     */
+    @Test
+    public void checkCallSetsRouting() throws Exception {
```
jeromekelleher: This seems something of an outlier. I'm expecting to see all the endpoints if CallSets are included here.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
```diff
@@ -0,0 +1,123 @@
...
+        testHttpMethods(makeFullUrl(partialUrl));
+    }
+
+    /**
+     * Test that searches' verbs/methods work as expected.
+     *
+     * @param fullUrl the URL to test
+     */
+    private void testSearchRouting(final String fullUrl) throws UnirestException {
+        // send some malformed requests and expect status == HTTP_BAD_REQUEST
```
jeromekelleher: This is a good test, but should be names something more specific. TestBadJson?

jeromekelleher:
I think the framework is good here and we should definitely have tests like this (we do need to check some top-level error behaviour and so on). However, a lot of the tests here don't make much sense to me. I would pare this down to just testing the BadJson on the search endpoints (for which the expected behaviour is pretty clear), and look to extend these tests over time in conjunction with updating the high-level documentation.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: Neither of those methods is a (JUnit) test: they're both utility methods called by test methods defined elsewhere in the file.

How to tell what's a test method from what's not?  Tests must be:
- annotated with the `@Test` tag.
- declared as `public void`.
- declared to throw some kind of exception, or subclass of exception.

Sorry if the name `testHttpMethods` led you to think otherwise.  If I recall correctly, older versions of JUnit looked for `public void` methods with names beginning with `test`, which probably contributed to the confusion.

The only true JUnit tests in this file are `checkCallSetsRouting`, `checkVariantSearchMethods`, and `checkSearchRouting`.

I'm not sure I understand your comment about multiple lines - can you explain?

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: Your point is well taken.  However, I wanted to provide some provenance for this test, since I couldn't find anything describing the behavior in ga4gh/schemas.

Perhaps it's not standard cross-server behavior at all, and this test should go away?

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: It's a POST sans payload, and the ref server returns 415 for that.  As above, maybe this is not something we want to require.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: Good point.  It's kind of vestigial.

It's easy enough to generalize this to run across all endpoints.  I will see if there's a reason not to do that.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: Yes, the present name is bad.  I'll come up with something more appropriate.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
jeromekelleher: Well I guess I was mildly complaining that I find it hard to follow stuff like
```java
 assertThat(Unirest.get(fullUrl).asBinary().getStatus()).isEqualTo(HttpURLConnection.HTTP_BAD_METHOD);
```
What's Unirest? Why do we call asBinary()? To me these are details distracting me from the point of the test, which is to check the http status. It would be easier to read if there was something like
```java
assertEqual(getHttpStatus(url), HttpURLConnection.HTTP_BAD_METHOD);
```
It seems to be the common style here, but I find it slow to parse these long chains of method calls, and generally resent having to put my finger on the screen to trace where brackets begin and end.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
jeromekelleher: I think if it's not currently documented (and not obviously the right thing to do) we should remove the test. Opening an issue on it for the schemas so that it is documented, and incorporating the test afterwards is the way to go I think. The baseline should be uncontroversial stuff, and we don't want undocumented assumptions made about the HTTP behaviour becoming requirements.

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
jeromekelleher: Again, if it's not documented we shouldn't require the behaviour. (I'm not saying it's not the right behaviour, just that it's not reasonable to make undocumented compliance requirements.)

cts-java/src/test/java/org/ga4gh/cts/api/endpoints/RawEndpointsIT.java
hjellinek: I hope the code I write is generally less dense than that, but I take your point.

jeromekelleher:
Looks good @hjellinek, thanks for the updates.

+1.

calbach:
+1

jeromekelleher:
Looks like everyone is happy with this; merging. Thanks @hjellinek and @macieksmuga!

PR #106, opened 2015-09-30
Fix for #88.

hjellinek:
Now this only tests what happens when you POST a bad payload to a search endpoint.
Also added a couple of missing search endpoints.


jeromekelleher:
Is this an update to the #105 PR? I think we should just update the PR directly.

Issue #75, opened 2015-09-09
compliance redux: expected layout of the compliance data is not described


jeromekelleher:
The expected layout of the test data in terms of reference sets, readgroupsets, variant sets etc is not described. There should be a section in the README_COMPLIANCE_DATA.md (which could perhaps be renamed to README.md since it's in the test-data directory?) which describes the expected data layout so server implementers know how they are supposed to import it.

It's perfectly reasonable to say that (e.g.) we require a dataset with name equal to "compliance_dataset_1" must contain exactly one variantset with name "variantset1" containing the variants in the VCF file X. Similar instructions should be given for readgroupsets and referencesets.

macieksmuga:
Makes sense to me. I'm on it.

calbach:
I'll piggyback on this issue since it's related. Let me know if you'd prefer I file separately:
- Ref and refset metadata is not mentioned, yet the tests require particular values for source accession.
- I found the docs for the BRCA1 reference region confusing. I initially thought that the tests expected the region to be loaded at the provided chr17 offset. I had to look at the test code to confirm that it was querying offset 0-N (not a position within 41196312:41277500).

Also +1 to renaming to README.md, so it shows up in GitHub automatically upon viewing the directory (I believe).

macieksmuga:
I believe this is now resolved via the merge of #112. @calbach - If you think there are still outstanding issues, I'll be happy to address them here.

calbach:
Looks great, thanks!

One small issue: looks like the SAMs have a `@SQ	SN:hs37d5` header. However, there's no corresponding reference in the reference set. Google's implementation would not allow you to associate the test reference set with this read group set, since the headers are incompatible.

macieksmuga:
Thank you for catching that, @calbach! I'll fix this immediately. Sorry for merging before you had a chance to get back to me about this.

PR #112, opened 2015-10-05
Updated reference set with JSON sidecar, adjusted variant metadata.

macieksmuga:
Consistency tweak and additional metadata to allow for running compliance data with the reference server.


macieksmuga:
@jeromekelleher @dcolligan could you take a look at this? Based on my testing, this PR is what's needed to get the compliance data working with the reference server.

hjellinek:
Note: this is related to the [Added shell script to wrangle compliance data into server format](https://github.com/ga4gh/server/pull/656) pull request.

calbach:
Per #75, what about the `Reference` metadata?

macieksmuga:
The main problem with the way this test-data directory is currently arranged is that it's not easy to expand to multiple datasets or reference sets. Thus, I was thinking of restructuring it into a more hierarchical directory format, possibly mirroring the reference server setup as follows:

```
- README_COMPLIANCE_DATA.md
- references/
  |- hg37.json
  |- hg37/
     |- ref_brca1.json
     |- ref_brca1.fa
- dataset1/
  |- <variant_set1>/
     |- <variants_file1>.vcf
     |- ...
  |- <variants_set2>/
     |...
  |- <reads_file1>.sam
  |...
- dataset2/
  |...
```

Another option would be to include an extra file (plaintext or json) that specifies what datafile goes with which reference or data-set.

macieksmuga:
@calbach The `Reference` metadata is what's contained in the two .json files (one already in Master, the other introduced as part of this PR). But it's not enough, as it doesn't affiliate the data files with specific named reference or data sets.

macieksmuga:
On second thought, it would make a lot more sense to just rename the files in `test-data` as follows:
```
brca1_<read_filename>.sam 
brca1_1kg-phase3_<variant_filename>.vcf
```
to make associations with datasets explicit, and assume that we're using a single reference set (named hg37 for the purposes of this dataset) for now.

this will be a) readable, b) trivial to implement and c) make it easy for me or others to write automatic data-bundlers for the various servers sourcing this data.

jeromekelleher:
Looks good to me @macieksmuga. I think the hierarchical formatting is probably the easiest approach, as it makes it pretty clear which dataset reads and variants belong to, and what the names of the various objects are supposed to be.

macieksmuga:
Thanks @jeromekelleher - I'm leaning towards the naming convention approach for now, as it'll be an incrementally smaller change. Going from that to a directory-hierarchical scheme (should we decide on going there) would also be relatively easy.

For context, I'm doing this now to fulfill @dcolligan's request for a python-based data prep script in https://github.com/ga4gh/server/pull/656 - and I realized that I should probably make it generic enough so that it didn't need to be adjusted every time I added another variant or read file (or dataset, for that matter).

macieksmuga:
Could we work to merge this ASAP? - the compliance data were updated here to fix small internal inconsistencies, and to add in necessary "sidecar" metadata json files.

jeromekelleher:
+1.

macieksmuga:
I'm about to squish commits & merge this PR, as it fixes a critical bug with compliance data (without it, https://github.com/ga4gh/server/pull/656 - already merged - won't work properly).

PR #127, opened 2015-11-09
Fix for bug #120.

### LINKED ISSUE #120: Change the logging level of the "set urlRoot" message from AvroJson
Change the logging level of the

```
org.ga4gh.ctk.transport.avrojson.AvroJson - set urlRoot = http://localhost:8000 path = ...
```

message from `INFO` to `DEBUG`.

hjellinek:
Fixed #120.
While in that file, I also made the urlRoot instance var private for consistency with the others.


calbach:
+1

macieksmuga:
+1

jeromekelleher:
Two +1s, so mergable.

PR #134, opened 2015-12-09
fixed #133

### LINKED ISSUE #133: Reference accession field should be an array as opposed to single
Was expecting a single value for the reference accessions, when in fact it should allow for multiple values.

screen shot 2015-12-09 at 11 50 57 am

macieksmuga:
Fixed issue #133 : Compared returned reference accessions against single element array as opposed to string.


david4096:
+1

david4096:
+1

calbach:
+1

sarahhunt:
+1

Issue #131, opened 2015-12-08
convert_to_binary.sh points to non-existent file


david4096:
On line 24 the hardcoded conversion points to a file that's not there.

https://github.com/david4096/compliance/commit/f4191ecea4416d49206f450f75a81c61372c6597

PR #135, opened 2015-12-09
Filename in convert_to_binary.sh is now correct.

macieksmuga:
Fixes issue #131. Also added a short preamble indicating that this script is for illustrative purposes only.


david4096:
+1

sarahhunt:
+1

PR #128, opened 2015-11-09
Utils.getValidReferenceId no longer assumes a single reference set or reference

hjellinek:
`Utils.getValidReferenceId` no longer assumes a single reference set or reference, but rather uses the IDs/names provided in `TestData` to look them up.

Removed unnecessary calls to it from `ReadGroupSetsPagingIT.checkTwoSimultaneousPagingSequencesThroughReadGroupSets` and `ReadsSearchIT.searchReadsProducesWellFormedReads`.

Fix for #117.

PR #136, opened 2015-12-09
Remove unecessary getReferenceId calls and check that reference set lists are length 1

### LINKED ISSUE #117: ReferenceBasesPagingIT should not assume there's a single reference
The tests in `ReferenceBasesPagingIT` should not assume there's a single reference.  There may be several.

Find and fix other places where tests make this assumption.

david4096:
Fix #117 by reproducing Herb's changes on PR #128. Sets the checksum for a SearchReferenceRequest, removes unecessary calls to getReferenceId, and tests to ensure the length of the reference set arrays are equal to 1.

### DIFF
--- a/cts-java/src/test/java/org/ga4gh/cts/api/Utils.java
+++ b/cts-java/src/test/java/org/ga4gh/cts/api/Utils.java
@@ -117,20 +117,16 @@ public class Utils {
      * @throws AvroRemoteException is the server throws an exception or there's an I/O error
      */
     public static String getValidReferenceId(Client client) throws AvroRemoteException {
-        final SearchReferenceSetsRequest refSetsReq = SearchReferenceSetsRequest.newBuilder().build();
-        final SearchReferenceSetsResponse refSetsResp = client.references.searchReferenceSets(refSetsReq);
-
-        final List<ReferenceSet> refSets = refSetsResp.getReferenceSets();
-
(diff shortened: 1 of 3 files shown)


cts-java/src/test/java/org/ga4gh/cts/api/Utils.java
```diff
@@ -117,7 +117,10 @@ public static String randomName() {
...
      */
     public static String getValidReferenceId(Client client) throws AvroRemoteException {
-        final SearchReferenceSetsRequest refSetsReq = SearchReferenceSetsRequest.newBuilder().build();
+        final SearchReferenceSetsRequest refSetsReq =
+                SearchReferenceSetsRequest.newBuilder()
+                                          .setAssemblyId(TestData.REFERENCESET_ASSEMBLY_ID)
+                                          .build();
         final SearchReferenceSetsResponse refSetsResp = client.references.searchReferenceSets(refSetsReq);
 
         final List<ReferenceSet> refSets = refSetsResp.getReferenceSets();
```
calbach: Should probably assert size 1 here.

calbach: Actually I would just replace this whole refset section with a call to the other helper:
`getReferencedSetIdByAssemblyId`

calbach:
+1 less a minor issue

macieksmuga:
+1

Issue #31, opened 2015-08-13
compliance_redux: update version number in POMs from 0.5.1-SNAPSHOT


hjellinek:
All of the `pom.xml` files that mention a version mention "`0.5.1-SNAPSHOT`".  We will need to change that, though I don't know what the new version will be yet.

diekhans:
Is it possible to have this in one file?  oh, XML entities ... never
mind

hjellinek:
Surprisingly, I believe it is possible: `parent/pom.xml`, in the `<properties>` section.  We _may_ be able to get Maven to use `${ga4gh.schema.local.version}` in the other places the string appears now.

    <properties>
        <ga4gh.schema.local.version>0.5.1-SNAPSHOT</ga4gh.schema.local.version>
        ...
    </properties>

PR #138, opened 2015-12-14
Changed internal version from 0.5.1-SNAPSHOT to 0.6.0a1

macieksmuga:
I should have caught this one before. The compliance suite was using the `0.5.1-SNAPSHOT` name internally all over the place. Please take a quick look and +1 so we can release this ASAP.

### DIFF
--- a/ctk-cli/pom.xml
+++ b/ctk-cli/pom.xml
@@ -23,13 +23,13 @@
     <description>CLI app to drive tests of GA4GH data server v0.5.1</description>
 
     <artifactId>ctk-cli</artifactId>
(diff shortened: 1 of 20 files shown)


diekhans:
+1 (I assume you grepped all the files)

Please create a ticket to check version to be set in one place and auto-substitured

david4096:
+1 I think it goes with this ticket: https://github.com/ga4gh/compliance/issues/31

macieksmuga:
Since this is a trivial syntactic change, I merged in preparation for release.

PR #141, opened 2015-12-17
Variant Annotation - initial version

sarahhunt:
Initial variant annotation tests & test data


david4096:
Hi Sarah, this looks good, thanks!

I've pulled down the PR and all the tests run great! I'm currently inspecting the documents that will be communicated to the server under test to make sure they adhere to the schema.

david4096:
@sarahhunt It seems prudent to have a variant for each of the annotations in the compliance dataset as well, doesn't it?

david4096:
@sarahhunt I'm trying to figure out why the ref server is returning 11 instead of 10 Variant Annotations from the WASH7P file. I just want to make sure our test is right even if server is returning something unexpected.

I've attached the response JSON for this request:

http post localhost:8000/variantannotations/search variantAnnotationSetId="YnJjYTE6MWtnUGhhc2UzOnZhcmlhbnRhbm5vdGF0aW9ucw==" referenceName=1 start:=10177 end:=11008

It's possible that the range query isn't exclusive on the `stop` coordinate since that last result appears to be from 11008.

[variantAnnotations.txt](https://github.com/ga4gh/compliance/files/70333/variantAnnotations.txt)

sarahhunt:
Hi @david4096!

**Thanks for finding this one - it is is an error in the compliance 
suite. My implementation initially had a page size limit of 10, hence 
the count!

I've pushed a PR though to fix it - #156 .

sarahhunt:
@david4096 - are you suggesting a GET /variants/{id} check that the ids 
in the annotation record return variant records? If so, I would agree - 
and the same for features once they are added via the RNA PR.

david4096:
@sarahhunt: Yes, having a variant for each annotation makes sense to me.

sarahhunt:
I've added the check on variantID in #158. It also impacts VariantSetsSearchIT, but the change to paging through 2 VariantSets rather than one is an improvement.

sarahhunt:
Additional information for Analysis, as discussed above, is added to the test data in #162. 

Also added is HGVSg annotation and variant names to allow improved checking of Variant records returned using the VariantAnnotation variantId.

sarahhunt:
Hi @david4096! There are extra Analysis tests, as you suggested in #163.

david4096:
@sarahhunt I've updated this branch with master so the merge should be easy :).

sarahhunt:
Thanks @david4096!

diekhans:
+1

david4096:
Latest commit updates this branch to be in line with schemas/master.

PR #147, opened 2015-12-18
Not all VariantSets have CallSets Fix #145

### LINKED ISSUE #145: Update CallsetsSearchResponseCheckIT to allow for VariantSet without CallSets
The test searchForExpectedCallSets assumes all VariantSets have CallSets, which is not required by the schema and no longer the case.

The opposite check, that a VariantSetId listed in a CallSet returned data, would be valid.

david4096:
Not all VariantSets have CallSets so this test has been changed to check that if a CallSet is returned, it contains reference to the VariantSet it came from. fix #145


macieksmuga:
+1

PR #148, opened 2015-12-18
New test for finding Variant Sets related to a Call Set fix #146

### LINKED ISSUE #146: new test: Find a VariantSet by its ID from a CallSet
As mentioned by @sarahhunt, we ought to be able to show the VariantSet's can be found by the list of VariantSetId's listed in a CallSet.

david4096:
Each CallSet contains a list of VariantSets. This test makes sure that VariantSets are found for each variantSetId in a CallSet record. fix #146

### DIFF
--- a/cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsSearchResponseCheckIT.java
+++ b/cts-java/src/test/java/org/ga4gh/cts/api/variants/CallSetsSearchResponseCheckIT.java
@@ -24,7 +24,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  */
 @RunWith(JUnitParamsRunner.class)
 @Category(VariantsTests.class)
-public class CallsetsSearchResponseCheckIT implements CtkLogs {
+public class CallSetsSearchResponseCheckIT implements CtkLogs {
 
     private static final URLMAPPING urls = URLMAPPING.getInstance();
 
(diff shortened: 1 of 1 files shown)


macieksmuga:
+1, agree with the test logic proposed here.

cts-java/src/test/java/org/ga4gh/cts/api/variants/CallsetsSearchResponseCheckIT.java
```diff
@@ -56,7 +56,41 @@ public void searchForExpectedCallSets() throws AvroRemoteException {
             assertThat(csResp.getCallSets()).isNotEmpty();
         }
     }
+    /**
+     * Make sure VariantSet records exist for each CallSet.
+     *
+     * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void searchForExpectedVariantSetsByCallSets() throws AvroRemoteException {
```
macieksmuga: One minor quibble: The name of this test makes it sound like you're searching for variant sets. Instead, you're using a variant set as the query to search for call sets, then verifying that these call sets contain that variant set. Wouldn't something like `testSearchCallSetsByVariantSet` be more appropriate?

david4096:
@macieksmuga Got the rename :)

PR #150, opened 2015-12-18
Fix broken test searchReadsWithUninterestingRangeProducesZeroReads

david4096:
Changed the test to match the title, no reads should be found in the range. We need to edit the range values to find a true empty range.

### DIFF
--- a/cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
+++ b/cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
@@ -38,34 +38,23 @@ public class ReadsSearchIT implements CtkLogs {
      * @throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
      */
     @Test
-    public void searchReadsWithUninterestingRangeProducesZeroReads() throws AvroRemoteException {
-
+    public void searchRangeWithNoReadsReturnsZeroResults() throws AvroRemoteException {
         final String refId = Utils.getValidReferenceId(client);
 
-        final long emptyRangeStart = 150;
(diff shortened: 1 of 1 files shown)


cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -39,33 +39,22 @@
      */
     @Test
     public void searchReadsWithUninterestingRangeProducesZeroReads() throws AvroRemoteException {
```
macieksmuga: Didn't Sherlock Holmes have something to say about a dog that didn't bark? ie. it's not that the range is uninteresting (it may be fascinating!) it's just that the dataset we're testing has no reads overlapping it, right? So, perhaps a new name is in order:

`searchRangeWithNoReadsReturnsZeroResults()`

cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
```diff
@@ -39,33 +39,22 @@
...
-
         final String refId = Utils.getValidReferenceId(client);
 
-        final long emptyRangeStart = 150;
+        final long emptyRangeStart = 150; // is this range actually empty?
```
macieksmuga: looking into this now...

macieksmuga:
I can't argue with the merits of this PR, but there are a couple of details I'm checking on - to see if the specified range is indeed free of reads, for example...

david4096:
@macieksmuga I changed the range to be 0,100... I wasn't finding any reads there but I'd hate to overfit to the ref server if this doesn't make sense. Test title should make more sense now too!

PR #151, opened 2015-12-19
sync schemas - use approved metadata records & correct feature search key

sarahhunt:
Hi @david4096

This updates the compliance schemas to those used by the schema & server repos.

I'm sending it as a PR so you can see the changes and merge when convenient. The main difference is the metadata fix, which should not have any impact on the tests.
In addition, SearchVariantAnnotationsRequest has had feature_ids renamed to featureIds to match the standard naming convention. The reference server and ensembl server have been updated for this change, so the compliance test filtering on feature id will fail until this PR is merged.


david4096:
Thanks Sarah +1, I pulled it down and didn't notice anything out of sorts!

PR #153, opened 2015-12-23
Changed ctk-domain pom.xml to delete generated assertions fix #152

### LINKED ISSUE #152: maven clean isn't removing ctk-domain auto generated assertions
Builds might fail after bringing in new schemas since assertion files are created by the build, but not completely removed.

To reproduce: `git checkout variant_annotation` and then `mvn clean install`. Then switch back to a branch with the older schemas `git checkout master` and run `mvn clean install` to see build failure.

david4096:
When switching back and forth from branches with different schema the list of assertion files to generate wasn't being deleted. Now it is, so you can `mvn clean` between schema changes without an issue.

### DIFF
--- a/ctk-domain/pom.xml
+++ b/ctk-domain/pom.xml
@@ -55,6 +55,18 @@
     <build>
         <plugins>
             <plugin>
+                <artifactId>maven-clean-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


macieksmuga:
+1 Very necessary.

PR #157, opened 2016-01-04
Added a test for finding variant annotation sets by their ID. fixed #144

### LINKED ISSUE #144: GET VariantAnnotationSet test
A test of the /variantannotationsets/{id}  endpoint is needed. As ids are server-specific, an id will need to be extracted using /variantannotationsets/search to pull out a VariantAnnotationSet. The two records can then be compared.

david4096:
Ref server is currently failing this test without the GET endpoint implemented.

### DIFF
--- a/cts-java/src/test/java/org/ga4gh/cts/api/variantAnnotation/VariantAnnotationSetsSearchIT.java
+++ b/cts-java/src/test/java/org/ga4gh/cts/api/variantAnnotation/VariantAnnotationSetsSearchIT.java
@@ -56,4 +56,21 @@ public class VariantAnnotationSetsSearchIT implements CtkLogs {
                              .forEach(vas -> assertThat(vas.getVariantSetId()).isNotNull());
 
     }
+
+    /**
+     * Check to see if variant annotation sets can be gotten by their ID.
+     *
+     *@throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkGetVariantAnnotationSetById() throws AvroRemoteException {
+        // Find a variant annotation set ID using search
+        final String variantAnnotationSetId = Utils.getVariantAnnotationSetId(client);
+
+        // Get a variant annotation set using GET
+        final VariantAnnotationSet vSet = client.variantAnnotations.getVariantAnnotationSet(variantAnnotationSetId);
+
+        assertThat(vSet).isNotNull();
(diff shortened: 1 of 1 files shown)


cts-java/src/test/java/org/ga4gh/cts/api/variantAnnotation/VariantAnnotationsSearchIT.java
```diff
@@ -140,6 +140,23 @@ public void checkTranscriptEffects() throws AvroRemoteException {
     }
 
     /**
+     * Check to see if variant annotations can be gotten by their ID.
```
sarahhunt: variant annotations => variant annotation sets

cts-java/src/test/java/org/ga4gh/cts/api/variantAnnotation/VariantAnnotationsSearchIT.java
```diff
@@ -140,6 +140,23 @@ public void checkTranscriptEffects() throws AvroRemoteException {
...
+     *
+     *@throws AvroRemoteException if there's a communication problem or server exception ({@link GAException})
+     */
+    @Test
+    public void checkGetVariantAnnotationsById() throws AvroRemoteException {
+        // Find a variant annotation set ID using search
+        final String variantAnnotationSetId = Utils.getVariantAnnotationSetId(client);
+
+        // Get a variant set using GET
```
sarahhunt: variant => variant annotation set

sarahhunt:
Hi @david4096 
This one looks good as a test of the GET variantannotationsets/{id} endpoint. 
The comments are a little confusing (as highlighted) and it would be clearer to call it something like 'checkGetVariantAnnotationSetsById'. It may be better in a seperate file too (VariantAnnotationSetsGetByIdIT.java?) as the existing file names all suggest checks on /search endpoints. 

I could also see a point to an additional similar test in VariantAnnotationsSearchIT.java to check the consistency of the VariantAnnotation record. It would pull out a few variant annotation records, extract the variantAnnotationSetId string for each and do a GET look up to check it isn't garbage.

david4096:
Hi @sarahhunt! The fetch by ID tests for all the other endpoints exist within the `*SearchIT` files for the other endpoints, so we can leave them for now. I moved the test to the appropriate file and fixed the comments, thanks!

sarahhunt:
That's clearer now!

PR #158, opened 2016-01-05
Check that Variant in the VariantAnnotation record exists 

sarahhunt:
This requires a second VariantSet, so VariantSetsSearchIT is also updated - it is a more useful check to page through more than one record.


david4096:
+1

PR #165, opened 2016-02-02
New test for effect search, update schemas

david4096:
Ontology terms changed since the last PR, this update brings it in line with those changes. In addition adds a test for filtering variant annotations using the transcript effect term.

This PR will merge in changes from master, which will remove some merge conflicts from #141.


cts-java/src/test/java/org/ga4gh/cts/api/variantAnnotation/VariantAnnotationsSearchIT.java
```diff
@@ -256,6 +251,40 @@ public void checkVariantAnnotationVariants() throws AvroRemoteException {
             Variant v = client.variants.getVariant(variantId);
             assertThat(v).isNotNull();
             assertThat(v.getId()).isEqualTo(variantId);
+            assertThat(v.getStart()).isEqualTo(start);
```
sarahhunt: Hi @david4096! This assert on start is failing for me - expected:<10[177]L> but was:<10[234]L>

VCF is 1-based but GA4GH is 0-based, so the variant at 10177 in the VCF is at 10176 in GA4GH terms and outside the target region which starts at 10177. The first variant returned is therefore at VCF 10235 or GA4GH 10234 rather than at the search region start.

david4096: Hi to you @sarahhunt! This was actually my mistake, the variant's `.start` shouldn't be equal to the query, as was asserted, just greater than what we requested:

    assertThat(v.getStart()).isGreaterThanOrEqualTo(start);

sarahhunt:
Looks good! +1

PR #166, opened 2016-02-02
protobuf adjustment of compliance suite

msackman:
PR for WIP.


ctk-transport/src/main/java/org/ga4gh/ctk/transport/protobuf/Base.java
```diff
@@ -0,0 +1,110 @@
...
+            } else {
+                String json = response.getBody().toString();
+                Common.GAException.Builder exceptionBuilder = Common.GAException.newBuilder();
+                try {
+                    JsonFormat.parser().merge(json, exceptionBuilder);
+                } catch (InvalidProtocolBufferException e) {
+                    log.warn("Failure when processing GAException: BODY <" + json + ">: ", e.toString());
+                    throw e;
+                }
+                log.info("Throwing GAException for " + json + ", status " + httpStatus);
```
cstorey: Does this want to be at error, with a more descriptive message?

msackman: Well, basically, I'm not sure. That was taken from the original (https://github.com/ga4gh/compliance/blob/master/ctk-transport/src/main/java/org/ga4gh/ctk/transport/avrojson/AvroJson.java#L247) it's just that my refactoring of this area is so extensive that diff is no where near smart enough.

I'm slightly dubious of the value of logging before throwing an exception anyway, so I wouldn't object to seeing all of these go. Alt, they should probably all be at the same level - warn perhaps?

ctk-transport/src/main/java/org/ga4gh/ctk/transport/protocols/Client.java
```diff
@@ -112,16 +126,14 @@ public SearchDatasetsResponse searchDatasets(SearchDatasetsRequest request)
...
-        public Dataset getDataset(String id) throws AvroRemoteException {
-            String path = urls.getGetDataset();
-            Dataset response = new Dataset();
-            final AvroJson aj =
-                    new AvroJson<>(response, urls.getUrlRoot(), path, wireTracker);
-            response = (Dataset) aj.doGetResp(id);
-            return response;
+        public Dataset getDataset(String id) throws InvalidProtocolBufferException, UnirestException, GAWrapperException {
+            Dataset.Builder responseBuilder = Dataset.newBuilder();
+            new Get<>(urls.getUrlRoot(), urls.getGetDataSet(), id, null, responseBuilder, wireTracker).performQuery();
```
cstorey: Do you need to create this every time?

msackman: Discussed IRL. Basically yes - the Base captures the entire lifecycle of the request, whether it's a Get or a Post.

msackman:
Ok, I think I'm done with this enough to turn it over to you. Let me know what you thoughts are.

david4096:
Hi @msackman I pulled this down and was able to build and run it. One thing I noticed is that requests aren't being formed properly, I expect the JSON payloads for `checkVariantsForExpectedReferenceName()` in VariantsSearchIT.java, for example, to maintain their type. The client builds `SearchVariantsRequest vReq` properly, but what is sent over the wire appears to have Numbers coerced to Strings.

The server received
```
{u'start': u'50', u'variantSetId': u'YnJjYTE6MWtnUGhhc2Uz', u'end': u'100', u'referenceName': u'ref_brca1'}
```

When it expected:
```
{u'start': 50, u'variantSetId': u'YnJjYTE6MWtnUGhhc2Uz', u'end': 100, u'referenceName': u'ref_brca1'}
```

Note the lack of quotes around the start and end values.

I did a quick experiment to see if this was due to google's protobuf JsonFormat by comparing the results of `gson.toJson(vReq)` with `JsonFormat.printer().print(vReq);` which underlies the client conversion.

```
// gson.toJson(vReq)
{"bitField0_":0,"variantSetId_":"YnJjYTE6MWtnUGhhc2Uz","callSetIds_":[],"referenceName_":"ref_brca1","start_":50,"end_":100,"pageSize_":0,"pageToken_":"","memoizedIsInitialized":1,"unknownFields":{"fields":{}},"memoizedSize":-1,"memoizedHashCode":0}

// JsonFormat.printer().print(vReq);
{
  "variantSetId": "YnJjYTE6MWtnUGhhc2Uz",
  "referenceName": "ref_brca1",
  "start": "50",
  "end": "100"
}
```

Note that com.google.gson has not wrapped the start and end fields with quotes.

Is this a problem with protobuf's JSON serialization?

diekhans:
@calbach, comments?  This looks like a bug in ProtoBuf Java library.

calbach:
Ah, I failed to point this out as a wire incompatibility in my PR. This behavior is, however, consistent with the specification: https://developers.google.com/protocol-buffers/docs/proto3#json

My understanding is that the primary reasoning for representing an int64 as a string in JSON is that Javascript cannot losslessly represent an integer greater than 2<sup>52</sup> due to implementation as a 64 bit float. Our options are:

1. Accept this as a wire incompatibility.
2. Switch int64s -> int32 where possible. We would have to ensure this is a reasonable maximum value on a case-by-case basis. In many cases, I think it may not be sufficient, e.g. genomic coordinate I could certainly imagine exceeding a value of 2 billion.

calbach:
Also one more note on int64/uint64: the spec allows for clients to specify either number or string, i.e. the server should always output a string, but the client may send a number or string.

diekhans:
Ouch.  Since JSON is now considered language neutral, that is
really frustrating that protobuf optimizes for one language.


However, there are no known chromosomes over 1G, although
speculation that they exist.

  https://www.biostars.org/p/12560/

uint32 is probably very safe.

Although, if the the protobuf JSON parser handles the
conversions, it's a motivation to not parse by hand ;-)

pgrosu:
@diekhans What about this instance?

https://en.wikipedia.org/wiki/Marbled_lungfish

~p

calbach:
It is slightly annoying, but I'd probably be more frustrated by a spec which dropped bits of information opaquely in a widely used language such as Javascript, especially when a common use case for protobuf-JSON is certainly consumption by Javascript. Also this is not an official statement, just my understanding.

uint32 sounds a bit too close for comfort for me, but I will concede if there is consensus. For now I will update the PR description to say that int64s are strings.

diekhans:
Those whole genome duplications are fun.  I haven't seen
what the maximum chromosome size is for the Marbled lungfish,
or any karyotype information.

diekhans:
CH Albach <notifications@github.com> writes:
> It is slightly annoying, but I'd probably be more frustrated by a spec which
> dropped bits of information opaquely in a widely used language such as
> Javascript, especially when a common use case for protobuf-JSON is certainly
> consumption by Javascript.

Having it wrong in every language is worse.  We simple can't
have numbers as strings.  This makes all the common use cases
with chromosome coordinates more complex.

> Also this is not an official statement, just my understanding.

thanks for educating me

> uint32 sounds a bit too close for comfort for me, but I will concede if there
> is consensus. For now I will update the PR description to say that int64s are
> strings.

BAM representation is 32bit.  There are no known chromosomes
that exceed signed 32bit ints.  As much as I love comparative
genomics, we can't go forward with numbers as strings.

msackman:
So I don't want to wade into this as I know nothing about genomics. However, a few things are interesting here:

1. When I turned this over, all the tests against the server were passing. That suggests that the python protobuf3/json is lenient in this way too - it must be accepting strings-as-numbers. Either that or something way more broken/weird was going on for me.
2. My initial reaction when I saw this issue be reported last night was "ahh, that makes sense, this is the JavaScript-can't-do-integer-maths issue". However, actually checking json.org, there is no suggestion numeric representation needs be constrained by any sort of numeric type - in JSON it appears you are free to directly encode numbers to any size and any precision.
3. So the issue is simply "what happens on the interpreting side?", and how to make that lossless, and can it be made lossless? If you don't need to do maths with these numbers at all in the browser, I'm wondering whether you can manually parse the JSON in javascript directly - it's just a string anyway, and thus keep numbers as strings there. Thus JavaScript is the only special case, which makes most sense as it's JavaScript that can't do integer maths. If you actually need to do maths on these numbers in the browser then you're sunk anyway - it's going to be hard to have confidence any of that maths will be right.
4. Actually just using the direct protobuf binary encoding may well be much much easier in the long run...

macieksmuga:
The really interesting point here, if I understand right, is given that the compliance suite itself is written using protobuf libraries, it effectively _can't_ test all the allowable forms of JSON that a server should accept. 

To wit: According to https://developers.google.com/protocol-buffers/docs/proto3#json, an int64 field (such as our `start` and `end` fields) should accept either number or string encoded inputs. However, the JSON generated by the compliance suite can only produce string values for this type, as it's internally using the protobuf->JSON libraries.

So, either we re-implement (again!) the compliance suite generating JSON directly - not relying on the protobuf libs - or we accept that there will be server requests acceptable per the schemas + protobuf specifications that just aren't being tested.

diekhans:
One could always have a small set of tests for a few cases of
different int64 encodings.  Doubtful that it is worth the cost
of testing every int64 case.

This could be done with a different encoder, maybe just post
process the JSON.

We will also have to test the binary protobuf encoding, so going
back to printing JSON isn't going to help with that.

david4096:
In the schemas it states that Avro JSON serialization libraries are used. We don't need to test all the edge cases if we state the same here. https://github.com/ga4gh/schemas/blob/master/doc/source/appendix/json_intro.rst

I would like to suggest that we simply require that Google's serialization is used.

macieksmuga:
Agree with @david4096 - my point was mostly of theoretical interest. @diekhans good point as well, although the current schemas explicitly specify JSON as the wire format, so compliance can't move over to binary until the schemas do.

irene-lshift:
We've just updated this PR with the latest changes from master.

macieksmuga:
I feel like I'm tripping over an old issue here, but I can't get several compliance tests to pass. 

The error pattern is always as follows:
 
`com.google.protobuf.InvalidProtocolBufferException: Expect message object but got: ["<some content here>"]`

For example for tests like `searchReadsWithOneReadGroupIdSucceeds`, I get the error: `com.google.protobuf.InvalidProtocolBufferException: Expect message object but got: ["60"]` 

FWIW, the alignments JSON object that I receive as a response from the server for the above test looks as follows:

```
{"alignments":[{"failedVendorQualityChecks":false,"improperPlacement":false,"fragmentLength":186,"alignedQuality":[15,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,37,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,40,37,37,37,37,37,37,33,37,37,37,33,33],"numberReads":2,"duplicateFragment":false,"secondaryAlignment":false,"readNumber":1,"readGroupId":"WyJicmNhMSIsInJncyIsIkhHMDAwOTYiLCJTUlIwNjI2MzQiXQ","fragmentName":"HG00096.mapped.ILLUMINA.bwa.GBR.low_coverage.20120522.bam.cram:117428991","alignedSequence":"CAGTCCTGGATAATGGGTTTATGAAAAACACTTTTTCTTCCTTCAGCAAGCAAAATTATTTATGAAGCTGTATGGTTTCAGCAACAGGGAGCAAAGGAAA","id":"WyJicmNhMSIsInJncyIsIkhHMDAwOTYiLCJIRzAwMDk2Lm1hcHBlZC5JTExVTUlOQS5id2EuR0JSLmxvd19jb3ZlcmFnZS4yMDEyMDUyMi5iYW0uY3JhbToxMTc0Mjg5OTEiXQ","alignment":{"mappingQuality":60,"cigar":[{"operationLength":"100","operation":"ALIGNMENT_MATCH","referenceSequence":""}],"position":{"strand":"POS_STRAND","position":"136","referenceName":"ref_brca1"}},"info":{"MQ":["60"],"RG":["SRR062634"],"MD":["100"],"X0":["1"],"X1":["0"],"SM":["37"],"AM":["37"],"XT":["U"],"NM":["0"]},"supplementaryAlignment":false,"nextMatePosition":{"strand":"NEG_STRAND","position":"222","referenceName":"ref_brca1"}}],"nextPageToken":""}
```
So either my server (that I believe is running the latest protobuf branch) is emitting stale JSON or the compliance suite is somehow being overly picky.

Any help with this would be greatly appreciated.

parent/pom.xml
```diff
@@ -41,8 +41,8 @@
 
         <ant.version>1.9.5</ant.version>
         <assertj.core.version>3.1.0</assertj.core.version>
-        <assertj.generator.version>1.6.0</assertj.generator.version>
-        <avro.version>1.7.7</avro.version>
+        <assertj.generator.version>2.0.0</assertj.generator.version>
+        <proto.version>3.0.0-beta-2</proto.version>
```
macieksmuga: Please update this to 3.0.0-beta-3 ASAP - it's the only thing that's holding up my +1

macieksmuga:
+1 modulo updating the protobuf version. I'll merge this as soon as it's updated.

irene-lshift:
I updated the protobuf version. I am a bit unclear about these failures, I saw them when running the tests on intellij, but they pass on the command line (with mvn clean install, and running the ./ctk script). One thing to keep in mind is that the server got a new protobuf version recently so might need to reinstall the dependencies and regenerate the compliance data.

macieksmuga:
+1 re-tested, works.

jeromekelleher:
+1.

PR #168, opened 2016-02-10
Check created time of analysis field

david4096:
Schemas were updated with ISO8601 time so we need to properly parse them in the analysis field. This compares a parsed expected result with the parsed server response.


david4096:
@sarahhunt Does this work for you?

sarahhunt:
Hi @david4096! Yes - this works for me; happy to merge.

PR #172, opened 2016-03-01
remove parent section and add back in old version

kozbo:
mvn clean install failed on my version of maven (3.3.9) until I made these changes. However mvn 3.3.3 builds correctly. I figure that we want to move forward with he latest version of maven.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,16 +33,9 @@
         <module>dist</module>
     </modules>
 
-    <parent>
-        <artifactId>ctk-parent</artifactId>
-        <groupId>org.ga4gh</groupId>
-        <version>0.6.0a2</version>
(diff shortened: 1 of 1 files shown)


macieksmuga:
+1 I checked this fix independently on my build, and it works as advertised. If anybody knows a reason to keep the `<parent>` section here, please speak up now!

david4096:
So I think we want to keep the new version number and the parent allows the submodules to understand their hierarchy. Since we're pretty carefully following how to structure the project using the filesystem we might be able to get away without it, but the real problem is that the version numbers didn't match across the .poms. I a quick search-replace for version number and everything went fine #173.

kozbo:
Pull request #173 supersedes this and is a more correct approach.

PR #173, opened 2016-03-02
bump version to 0.6.0a2

david4096:


kozbo:
+1 Replicated the success of this approach on my system.

PR #174, opened 2016-03-02
s/properPlacement/improperPlacement/

calbach:
Depends on merge of ga4gh/schemas#561.

Needs to be checked carefully. This was the only reference I was able to find to this field.


macieksmuga:
+1 Looks good to me. Since this change is already in schemas, I'm adding it in.

PR #175, opened 2016-03-04
Version bump, point to dataset

david4096:
With the latest server/master we have removed padding from identifiers. This fix brings the version number in line with schemas and removes `=` from datasetId.


david4096:
@macieksmuga fixed

PR #176, opened 2016-03-10
Sequence Annotations compliance starter batch

macieksmuga:


macieksmuga:
Removed WIP tag, as tests are now looking roughly like I want them to be.

PR #178, opened 2016-03-15
update checkSearchingAllReferencesReturnsWellFormed

### LINKED ISSUE #161: ReferenceSetsSearchIT expects all ReferenceSets to have the same md5 etc
ReferenceSetsSearchIT test checkSearchingReferenceSetsReturnsWellFormed seaches for all ReferenceSets and checks they match the test data (id, md5 and accessions). 

This test is fails for implementations serving more than just the test data.

sarahhunt:
The current test searches for all reference sets then checks they match the test data. This fails for servers with more than just the compliance data. 
This PR replaces the comparision to compliance data with checks that the id is not null and the md5 looks like an md5. 

Fixes #161


david4096:
+1

vadimzalunin:
:+1:

Issue #184, opened 2016-04-07
Test all read groups in a set search


david4096:
The current tests for reads search could be amended to cover the case when all the readgroups within an individual BAM are being searched.

[This test](https://github.com/ga4gh/compliance/blob/master/cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java#L185) checks across readgroupsets. Creating another test like this one that limits to only searching within a single readgroupset would demonstrate a useful feature.

dcolligan:
Fixed in #185

PR #185, opened 2016-04-14
Test readsSearch with multiple readGroups

dcolligan:
Issue #184

### DIFF
--- a/cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
+++ b/cts-java/src/test/java/org/ga4gh/cts/api/reads/ReadsSearchIT.java
@@ -17,6 +17,7 @@ import org.junit.experimental.categories.Category;
 import java.net.HttpURLConnection;
 import java.util.Collections;
 import java.util.List;
+import java.util.ArrayList;
 import java.util.stream.Collectors;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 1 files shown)


macieksmuga:
-1 
This test does not encompass the actual requirement as stated in the schemas (that queries across multiple read groups, whether or not they belong to the same read group set, succeed). That full requirement is already tested via `searchReadsWithAllIdsReturnsReadsForEach`.

That being said, I do like the actual contents of the test: That each returned alignment have its readGroupId contained in the set of readGroupIds queried. So on that strength alone, this test may be a valuable addition to the suite.

However, we've been trying to get away from overfitting compliance tests to the specific behaviors of the reference server. This test is specifically crafted to the known behavior of the reference server, and as such belongs in the reference server's suite of tests, not here.

diekhans:
I don't understand the objection.  This adds a legitimate test
cases for any server.  It does not remove any tests to make the
reference server happy.

As long as there is a test for readgroups across readgroupsets
and the reference server is failing it, this isn't over-fitting
the current restrictions.

david4096:
@macieksmuga understandable, fortunately we still have the other test in place and are failing appropriately. I asked for this test to be added since it covers what I believe to be a very common use case: "all the reads in a BAM". We won't forget that we're still under-implementing.

@dcolligan you could consider using list comprehension as opposed to a for-loop with something like this:

```
        final List<String> readGroupIds = readGroups
                .stream()
                .map(a -> a.getId())
                .collect(Collectors.toList());
```

I would like it if the comment stated this is for the "all the reads in a readgroupset" use case. "multiple read groups" is confusing and suggests that it might be testing some other case, like a subset of readgroups from a readgroupset. "Verifies that a request for all the readGroupIds in a Read Group Set returns only alignments with those readgroupIds." You might also improve the test by verifying some other fields of the alignment are well formed. +1

macieksmuga:
After talking it over with folks, +1 once @david4096's suggestions are incorporated (list comprehension and revised documentation).

May also be a good idea to include the _other_ common use case that Richard Durbin pointed out in a call (not currently supported by the reference server) for balance: The request for reads in a small genomic coordinate region (say, immediately surrounding a single known SNP), across ALL read groups.

dcolligan:
@david4096 updated

david4096:
thanks @dcolligan !

PR #186, opened 2016-04-14
Update default datasetid

david4096:
The ID scheme in the reference server has changed so the documentation and default datasetId is changed in this PR. @dcolligan @macieksmuga


macieksmuga:
+1 Tested, works.

PR #188, opened 2016-04-15
Changed expected server response on bad request to 400

### LINKED ISSUE #187: Status code for no IDs: readSearch
In SearchReads the expected status code for a request with no readgroupids is currently 501, not implemented. However the signature was changed in the reference server to return a 400. Which is correct?

`[TESTLOG] FAILED searchReadsWithNoIdsFails(org.ga4gh.cts.api.reads.ReadsSearchIT) due to expected:<[501]> but was:<[400]>`

macieksmuga:
Resolves #187


david4096:
+1 merging

PR #191, opened 2016-05-18
Add new test ontology data

david4096:
The ontology test data was in a custom file format. This adds the sequence ontology as the standard OBO file. https://github.com/ga4gh/server/pull/1225


macieksmuga:
+1
The new `obo` file has everything the old one had, and plenty more besides in a readable format. Even if we weren't doing the server ontology change, this would be a good idea.

PR #195, opened 2016-06-16
BioMetadata

david4096:
The biometadata API provides functionality to filtering data by sample in the Call Sets and Read Group APIs. It introduces the objects `BioSample` and `Individual`, with endpoints for searching and retrieving each.

Note that this PR is running against the `https://github.com/ga4gh/schemas/tree/biodataproto` branch to facilitate review. If the schemas PR is accepted then this branch will be updated to point at the proper commit hash.

The tests in this PR are based on metadata regarding the current test reads and calls. These data were generated from the thousand genomes dataset and represent realistic minimum expectations for tracking samples across data types.

There are also tests to ensure that the search-by-name and get-by-id functionality functions on both the biosamples and individuals endpoints operate as expected. BioSamples are further tested for the ability to be filtered by the individual ID they originate from.

For more details on the schemas see https://github.com/ga4gh/schemas/pull/636


cts-java/src/test/java/org/ga4gh/cts/api/TestData.java
```diff
@@ -148,6 +148,21 @@ private TestData() {
...
+
+    /**
+     * The VariantSet name in the test data.
+     */
+    public static final String VARIANTSET_NAME = "1kgPhase3";
+
+    /**
+     * The individual name in the test data.
+     */
+    public static final String INDIVIDUAL_NAME = "HG00096";
```
macieksmuga: We may want, as part of a larger overhaul of the compliance test dataset, to throw in a couple of individuals here, one with no biosamples, and one with multiple associated biosamples.

macieksmuga:
+1 Ran and verified that all works as expected. Tests look good too. Thanks @david4096!

cts-java/src/test/java/org/ga4gh/cts/api/TestData.java
david4096: https://github.com/ga4gh/compliance/issues/197

PR #207, opened 2016-08-03
Fix checkExpectedNumberOfRnaQuantifications

dcolligan:
@saupchurch is there really only supposed to be 1 rnaQuantification loaded into the server in the current test_data dataset?


saupchurch:
at this point there is a single rnaQuantification with several transcript quantifications for the test gene.  It probably would be a better set of tests if there were additional rnaQuantifications as part of the test rnaQuantificationSet

dcolligan:
ok could you give this a +1, then?

saupchurch:
can do! +1

PR #209, opened 2016-08-06
RNA Expression Level Tests

david4096:
This adds a few tests of the RNA functionality including filtering by feature IDs and threshold values. I believe these tests present the minimum level of request and response functionality.

A few more tests could be added that show that responses are well formed and contain values expected from the test data. @dcolligan @saupchurch


dcolligan:
+1

PR #211, opened 2016-08-10
RNA Update

david4096:
Fixes a capitalization bug in the callsets endpoint and adds a test ensuring that features exist for expression level objects. @dcolligan @saupchurch


dcolligan:
+1

PR #212, opened 2016-08-11
RNA Update

david4096:
After fixing thresholds to be non-inclusive, expression entries with 0.0 were being removed from results. This PR changes the number of expected results to reflect the default threshold behavior.

Column titles in the test data were altered to reflect their meaning in the server (names, not IDs).


saupchurch:
+1

PR #217, opened 2016-09-03
Check that version of dataset is set in info

### LINKED ISSUE #215: Test info on dataset

david4096:
Close #215


david4096:
Going to use https://github.com/ga4gh/compliance/pull/241 instead

PR #220, opened 2016-09-14
Unify properties for jitpack and version

david4096:
The important hand edited properties are now editable in one place (parent/pom.xml).

The schema version was being set in multiple places. Added a brief comment that explains how to switch to your own repo for schemas. Place the versioning in a single property so it is reused throughout.
Add docs for usage

Pushes version to 0.6.0a7 and adds version as property of parent/pom.xml.
@bwalsh

### DIFF
--- a/ctk-cli/pom.xml
+++ b/ctk-cli/pom.xml
@@ -23,13 +23,13 @@
     <description>CLI app to drive tests of GA4GH data server v0.5.1</description>
 
     <artifactId>ctk-cli</artifactId>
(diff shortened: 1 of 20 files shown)


bwalsh:
+1 thank you for the `docs/UpdatingTheSchemas.md`  document

david4096:
Depends on https://github.com/ga4gh/compliance/issues/224

dcolligan:
+1 after squash

PR #221, opened 2016-09-15
update g2p test data

bwalsh:
* bug fix:   update test-data/cgd-test.ttl to the latest from server/tests/data


david4096:
+1 thanks!

PR #222, opened 2016-09-19
Remove unused test data

david4096:
I mistakenly tried to add this data with the compliance dataset. It should be removed.

@saupchurch


saupchurch:
+1
no longer need these

PR #237, opened 2016-12-20
Use genotype list value

### LINKED ISSUE #225: Test for variant with call missing data
Add a test to cover the case described https://github.com/ga4gh/server/issues/1389 and reflects this schema change https://github.com/ga4gh/schemas/pull/735 . This may require modifying an existing VCF or adding a VCF that needs to interchange other types of call data than what is currently supported.

### LINKED ISSUE #236: Calls in vartiants/search not tested properly
After the switch to protobuf the compliance tests for calls did not make the request properly, and the tests did not reflect that no calls were being returned in checkCallsForWellFormedness.

david4096:
Closes #225 to bring us in line with https://github.com/ga4gh/schemas/pull/735 .

Also closes #236 by asserting some calls are returned.


kozbo:
Tested running compliance with correct results. +1

